Issue #41, opened 2020-02-11
[BUG] Cell detection fails on Windows


adamltyson:
Cell detection on windows fails with:

```python
C:\Users\adam.GARFIELD-WIN\.conda\envs\cellfinder\python.exe C:/Users/adam.GARFIELD-WIN/projects/cellfinder/cellfinder/main.py -s C:\Users\adam.GARFIELD-WIN\data\cellfinder\CT_BY_70_2_stitched\stitched\ch00 -b C:\Users\adam.GARFIELD-WIN\data\cellfinder\CT_BY_70_2_stitched\stitched\ch01 -o C:\Users\adam.GARFIELD-WIN\data\output\cellfindertest2 -x 2 -y 2 -z 5 --start-plane 1288 --end-plane 1305
2020-02-11 13:25:14 PM - INFO - MainProcess fancylog.py:265 - Starting logging
2020-02-11 13:25:14 PM - INFO - MainProcess fancylog.py:267 - Multiprocessing-logging module found. Logging from all processes
2020-02-11 13:25:14 PM - INFO - MainProcess main.py:64 - Skipping registration
2020-02-11 13:25:14 PM - INFO - MainProcess main.py:110 - Detecting cell candidates
Processing planes:   0%|          | 0/17 [00:00<?, ?it/s]Traceback (most recent call last):
  File "C:/Users/adam.GARFIELD-WIN/projects/cellfinder/cellfinder/main.py", line 158, in <module>
    main()
  File "C:/Users/adam.GARFIELD-WIN/projects/cellfinder/cellfinder/main.py", line 91, in main
    run_all(args, what_to_run)
  File "C:/Users/adam.GARFIELD-WIN/projects/cellfinder/cellfinder/main.py", line 112, in run_all
    detect.main(args)
  File "C:\Users\adam.GARFIELD-WIN\projects\cellfinder\cellfinder\detect\detect.py", line 114, in main
    bf_process.start()  # needs to be started before the loop
  File "C:\Users\adam.GARFIELD-WIN\.conda\envs\cellfinder\lib\multiprocessing\process.py", line 112, in start
    self._popen = self._Popen(self)
  File "C:\Users\adam.GARFIELD-WIN\.conda\envs\cellfinder\lib\multiprocessing\context.py", line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
[...]

PR #92, opened 2020-04-28
split setup function, cython classes called inside Mp3DFilter.process

EmanPaoli:
now seems to work on windows


adamltyson:
I had to add a couple of import statements, but this seems to work on linux and windows, thanks!

It's maybe not the most elegant solution, but I can't see it being much slower, so for now I'm closing #41.

Thanks again, much appreciated!

EmanPaoli:
happy to help!

PR #195, opened 2022-02-21
Avoid double runs on pr

paddyroddy:
Same as https://github.com/brainglobe/cellfinder-core/pull/33


paddyroddy:
Discussed that we don't want this behaviour as the benefits of checking thoroughly outweigh the extra CI minutes used

PR #197, opened 2022-03-14
Use relative paths for test data directories

dstansby:
This allows tests to be run in any directory (not just the root directory of the project).

### DIFF
--- a/tests/tests/conftest.py
+++ b/tests/tests/conftest.py
@@ -1,22 +1,13 @@
-import os
+import pathlib
 import sys
 
 import pytest
 from cellfinder_core.download.cli import main as cellfinder_download
 from imlib.general.config import get_config_obj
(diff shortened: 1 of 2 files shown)


tests/tests/conftest.py
```diff
@@ -1,22 +1,13 @@
...
-    "data",
-    "brain",
-)
-test_output_dir = os.path.join(
-    os.getcwd(),
-    "tests",
-    "data",
-    "registration_output",
-)
+test_data_dir = pathlib.Path(__file__) / ".." / ".." / "data"
```
paddyroddy, line 8: N.B. there's also the `parents[1]` option. But IMO this is more obvious what's going on

paddyroddy (approved):
Relative paths always

PR #205, opened 2022-05-19
Don't specify QT backend in requirements

dstansby:
See https://napari.org/plugins/best_practices.html#don-t-include-pyside2-or-pyqt5-in-your-plugin-s-dependencies as to why this is not a good thing to do.


paddyroddy (approved):
Always bothered me, this one. Think I have issues with `pyside2` on M1

Issue #207, opened 2022-05-27
[BUG] cellfinder interacts badly with multiprocessing_logging on Python 3.9.13


adamltyson:
@dstansby Nothing has changed with cellfinder, is this something to do with the actions?

paddyroddy:
It appears to be failing with this
```
Traceback (most recent call last):
File "D:\a\cellfinder\cellfinder\.tox\py39-\lib\site-packages\multiprocessing_logging.py", line 81, in _receive
  record = self.queue.get(timeout=0.2)
File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\multiprocessing\queues.py", line 113, in get
  if not self._poll(timeout):
File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\multiprocessing\connection.py", line 260, in poll
  self._check_closed()
OSError: [WinError 6] The handle is invalid
File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\multiprocessing\connection.py", line 335, in _poll
  return bool(wait([self], timeout))
File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\multiprocessing\connection.py", line 888, in wait
  ov.cancel()
File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\multiprocessing\connection.py", line 335, in _poll
  return bool(wait([self], timeout))
File "D:\a\cellfinder\cellfinder\.tox\py39-\lib\site-packages\multiprocessing_logging.py", line 81, in _receive
  record = self.queue.get(timeout=0.2)
OSError: [WinError 6] The handle is invalid
OSError: handle is closed
```
which I think is related to https://stackoverflow.com/questions/40108816/python-running-as-windows-service-oserror-winerror-6-the-handle-is-invalid

I'm a bit confused, though, as I don't see any use of `subprocess` or `multiprocessing`. Any idea @alessandrofelder ?

adamltyson:
There's multiprocessing in `cellfinder-core`?

dstansby:
Taking a look into this - will post thoughts/investigations as I do them.

Comparing a run that passed (https://github.com/brainglobe/cellfinder/runs/6576037299) and one that didn't (https://github.com/brainglobe/cellfinder/runs/6613252660) the only differences in dependencies installed are:

- pysygnal 0.3.4 > 0.3.5
- hsluv 5.0.2 > 5.0.3
- dask 2022.5.0 > 2022.5.1

[...]

PR #209, opened 2022-06-06
Remove multiprocessing-logging from requirements

dstansby:
The errors in https://github.com/brainglobe/cellfinder/issues/207 are being emitted by `multiprocessing_logging`. Since this package doesn't seem to be used anywhere in `cellfinder` itself, I've removed it here to see if it fixes the tests. I suspect it is being automatically invoked as a pytest plugin, and not interacting well with the multiprocessing being done in `cellfinder-core`.

@adamltyson is there a reason to keep `multiprocessing_logging` as a requirement?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@ requirements = [
     "tifffile",
     "natsort",
     "tqdm",
-    "multiprocessing-logging",
     "psutil",
     "configobj",
     "tensorflow>=2.5.0",


adamltyson:
Yeah, it's used by `fancylog` to ensure the logging works with the multiprocessing module.

dstansby:
👍  - I'm going to leave this running to see if `multiprocessing-logging` is the culprit, but will mark a draft so it doesn't get merged.

dstansby:
`multiprocessing-logging` was still pulled in as a requirement somewhere, but given it needs to stay I'm going to close this PR without merging.

PR #225, opened 2023-02-24
Clean requirements

dstansby:
- Remove un-used requirements
- Make list alphabetical

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,24 +7,20 @@ with open(path.join(this_directory, "README.md"), encoding="utf-8") as f:
     long_description = f.read()
 
 requirements = [
+    "brainreg",
     "cellfinder-core>=0.2.4",
+    "configobj",
+    "fancylog>=0.0.7",
+    "imio",
+    "imlib>=0.0.26",
+    "multiprocessing-logging>=0.3.4",
+    "natsort",
     "numpy",
-    "scikit-learn",
-    "configparser",
     "pandas",
     "packaging",
     "scikit-image",
     "tifffile",
-    "natsort",
     "tqdm",
-    "multiprocessing-logging>=0.3.4",
-    "psutil",
-    "configobj",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -7,24 +7,20 @@
...
     "pandas",
     "packaging",
     "scikit-image",
     "tifffile",
-    "natsort",
     "tqdm",
-    "multiprocessing-logging>=0.3.4",
-    "psutil",
-    "configobj",
-    "slurmio>=0.0.4",
```
adamltyson, line 23: `slurmio` _is_ used in some specific situations, but I agree it shouldn't be specified here. It's used in `cellfinder-core` [(`cellfinder` -> `cellfinder-core` -> `imlib`)](https://github.com/adamltyson/imlib/blob/625193be4a586d9040a48df9d51dbdd3a17c7d06/imlib/general/system.py#L137) to determine how many CPU cores to use in a SLURM-based HPC environment. 

We should support other schedulers in this way, but in the mean time, what do you think about a `[slurmio]` optional dependency?

dstansby: Given `slurmio` is used directly in `imlib`, I think the right thing to do here is to have an optional dependency section in `imlib`, and then cascade that up to the other libraries by including `imlib[slurmio]` in their depdencies.

As it stands `cellfinder` will still pull in `slurmio` via. the dependency path above, so it's probably safe to merge this PR as is, and discuss having an extra dependency section for `slurmio` in `imlib`?

adamltyson: Sounds good. I've raised an issue (https://github.com/adamltyson/imlib/issues/17). However generally, I think I'd like to move away from imlib, and either:
* Moving these utility functions into the relevant packages themselves (if there aren't many)
* Creating a `bg-utils` package if there are some heavily re-used functions. `imlib` is too general purpose at the moment.

PR #241, opened 2023-08-02
Update manual version

willGraham01:
Apologies for the PR noise.

New version in setup.py for patched version, so we can release `v0.7.0` to PyPI.

There look to be old tags (v0.6.0) that are datestamped 2020, but the latest releases are from 2021 and only reach v0.4.21.

Bumping to version `v0.7.0` to skip around this ambiguity. Will generate release notes once pushed to PyPI.

### DIFF
--- a/cellfinder/__init__.py
+++ b/cellfinder/__init__.py
@@ -1,3 +1,3 @@
-__version__ = "0.4.21"
+__version__ = "0.7.0"
 __author__ = "Adam Tyson, Christian Niedworok, Charly Rousseau"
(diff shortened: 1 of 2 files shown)


adamltyson:
@willGraham01 this package still uses [bump2version](https://github.com/c4urself/bump2version) to update versions (it hasn't get been migrated to setuptools_scm. I think `bump2version minor` should update the version in all the right places.

adamltyson:
If not, the version just also needs updating in `cellfinder/__init__.py`

adamltyson (approved):

PR #259, opened 2023-11-13
Replace `pandas.df.append` with `safe_concat`

### LINKED ISSUE #254: Deprecation of pandas.DataFrame.append() causing cellfinder to crash
Due to the [deprecation of `pandas.DataFrame.append()`](https://pandas.pydata.org/pandas-docs/version/1.4/reference/api/pandas.DataFrame.append.html), cellfinder crashes at [the analyse step](https://github.com/brainglobe/cellfinder/blob/21f724b46c84cf2c2489d1d4347c6c19f9277b45/cellfinder/analyse/analyse.py#L158).

This should be an easy fix, similar to those in other BG packages.

[Image.sc issue for reference](https://forum.image.sc/t/brainreg-cellfinder-on-ome-ngff-datasets/87853/5?)

alessandrofelder:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
Newer versions of `pandas` break the analysis code, due to the deprecation and removal of `pandas.Dataframe.append`.

**What does this PR do?**
* Replaces the use of `pandas.Dataframe.append` with `brainglobe_utils.pandas.safe_concat` in `analyse.get_region_totals()`
* Adds a regression test to check that the result is still the same as with `pandas` 1.5.3, ie before deprecation).

## References

Closes #254 

* Regression data generated with `pandas 1.5.3` and test written.

## Is this a breaking change?

Nope

Nope

## Checklist:

### DIFF
--- a/tests/tests/test_unit/test_tools/test_system.py
+++ b/tests/tests/test_unit/test_tools/test_system.py
@@ -14,12 +14,12 @@ background_im_dir = os.path.join(data_dir, "background")
 
 def test_get_subdirectories():
     subdirs = system.get_subdirectories(data_dir)
-    assert len(subdirs) == 9
+    assert len(subdirs) == 10
     assert Path(data_dir / "cells") in subdirs
     assert Path(data_dir / "brain") in subdirs
 
     subdir_names = system.get_subdirectories(data_dir, names_only=True)
-    assert len(subdir_names) == 9
+    assert len(subdir_names) == 10
     assert "cells" in subdir_names
     assert "brain" in subdir_names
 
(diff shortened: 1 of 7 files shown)


tests/tests/test_unit/test_tools/test_system.py
```diff
@@ -14,12 +14,12 @@
 
 def test_get_subdirectories():
     subdirs = system.get_subdirectories(data_dir)
-    assert len(subdirs) == 9
+    assert len(subdirs) == 10
```
willGraham01, line 17: Why is this now 10 and not 9? I can't see why this would have changed but I'm also not super knowledgeable about what `cellfinder` is meant to be doing here.

willGraham01 (approved):

tests/tests/test_unit/test_tools/test_system.py
alessandrofelder: This test checks the number of subdirectories in `tests/data` and I've added a new `analyse` subdirectory as part of the regression test introduced in the PR.

PR #260, opened 2023-11-14
Fix pathing problems

willGraham01:
@alessandrofelder Noticed your test data wasn't being found because our `MANIFEST` is told to ignore the `tests` directory when packaging.

Changed the `tox` command and paths in the to-be regression tests to point to the repository clone, which should then resolve the paths correctly using the repo clone.

### DIFF
--- a/tests/tests/test_unit/test_analyse.py
+++ b/tests/tests/test_unit/test_analyse.py
@@ -1,5 +1,5 @@
 import filecmp
-from importlib.resources import files
+import os
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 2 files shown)


willGraham01:
Looks like the tests are still failing, but it looks like a genuine test fail now? Rather than a path problem.

alessandrofelder:
I'll merge and fix on my branch (have already fixed one of the problems locally) - thanks!

willGraham01:
No problem - fingers crossed I haven't undone that local fix :sweat_smile:

PR #261, opened 2023-11-15
Prepare for migration

willGraham01:
See https://github.com/brainglobe/BrainGlobe/issues/45

Prepares the CLI-cellfinder package for migration into brainglobe-workflows.

Note that we do not want to archive the repository, since we need it for when we merge in cellfinder-core and cellfinder-napari.

On merge, tag new release version so that the deprecation warning files through to PyPI.

### DIFF
--- a/cellfinder/__init__.py
+++ b/cellfinder/__init__.py
@@ -1,7 +1,7 @@
 import warnings
 
 warnings.warn(
-    "cellfinder (CLI) has migrated. Please use brainglobe-workflows instead: https://github.com/brainglobe/brainglobe-workflows",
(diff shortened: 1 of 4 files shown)


cellfinder/__init__.py
```diff
@@ -1,3 +1,10 @@
+import warnings
+
+warnings.warn(
+    "cellfinder (CLI) has migrated. Please use brainglobe-workflows instead: https://github.com/brainglobe/brainglobe-workflows",
```
alessandrofelder, line 4:
```suggestion
    "cellfinder (the command line tool) has migrated. Please use brainglobe-workflows instead: https://github.com/brainglobe/brainglobe-workflows",
```
Just because some users may not know what a CLI is? Just a thought.

alessandrofelder (approved):

PR #348, opened 2024-01-03
Update tests

### LINKED ISSUE #271: Remove `test/data/brain` directory
I _think_ this isn't used, and was copied from the cellfinder repo. It was for registration testing.

adamltyson:
Removes some unnecessary test data and registers the custom `slow` pytest marker (this had been removed at some point in the migration). 

Closes #271


willGraham01 (approved):
Tests are happy locally as well, always like a negative diff :grin:

PR #349, opened 2024-01-03
Prep merge

### LINKED ISSUE #272: How are `prep_classification` and `prep_training` different?
In a PR discussion with @alessandrofelder, we noticed these two functions seem to only vary in the order of their input parameters:
- [prep_classification definition](https://github.com/brainglobe/cellfinder-core/blob/58141d7730dd852044440ba45b6e53702d8d0b63/src/cellfinder_core/tools/prep.py#L23C25-L23C25)
- [prep_training definition](https://github.com/brainglobe/cellfinder-core/blob/58141d7730dd852044440ba45b6e53702d8d0b63/src/cellfinder_core/tools/prep.py#L39)

If both classification and training require the same setup, maybe we can combine them in one function and name it `prep_tensorflow_and_model` or similar?

adamltyson:
Combines two very similar functions and also removes a seemingly redundant function arguement. 

@sfmig I've somewhat naively updated the benchmarking code, but I may have done this wrong.

Closes #272

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -127,7 +127,7 @@ jobs:
 
   build_sdist_wheel:
     name: Build source distribution and wheel
-    needs: [test, test_brainmapper_cli, test_numba_disabled]
+    needs: [test, test_numba_disabled]
     if: github.event_name == 'push' && github.ref_type == 'tag'
     runs-on: ubuntu-latest
     steps:
--- a/benchmarks/benchmarks/tools/prep.py
+++ b/benchmarks/benchmarks/tools/prep.py
@@ -4,10 +4,9 @@ from pathlib import Path
 from brainglobe_utils.general.system import get_num_processes
 
 from cellfinder.core.tools.prep import (
-    prep_classification,
+    prep_model_weights,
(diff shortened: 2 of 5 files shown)


adamltyson:
brainmapper updated in https://github.com/brainglobe/brainglobe-workflows/pull/66

willGraham01:
Worth flagging here that when this update and the matching one on `workflows` goes in, we should make new releases for both. (In particular, the `workflows` PR should force a minimum version of `cellfinder` that corresponds to the new release)

adamltyson:
> In the interest of resolving this, I guess we make the brainmapper CI check in this repo optional when creating a release and/or merging a PR? That way we at least have a check which'll indicate when things can potentially go wrong and need to be fixed in workflows too.

Yeah I think that's the way.

willGraham01:
> Yeah I think that's the way.

Have added this change to this branch.

willGraham01:
@adamltyson do you want me to merge this, publish the release, and then try re-running the CI on https://github.com/brainglobe/brainglobe-workflows/pull/66 ?

adamltyson:
Sounds good

willGraham01 (approved):

PR #351, opened 2024-01-03
Update tooling config

adamltyson:
- Remove references to py38
- Remove cibuildwheel config (unnecessary since the move to numba from Cython)
- Move tox config to pyproject.toml to reduce number of config files and harmonise with the rest of BrainGlobe


willGraham01 (approved):
`tox` seems happy when run locally, so approving :+1: (Had to merge in main due to another update)

willGraham01 (approved):

PR #365, opened 2024-01-16
Delete old training data saves and warn user

adamltyson:
- Deletes old training data when resaving to ensure that only the correct data is saved
- Warns users about this
- Tidies up the code base by using qtpy functions from brainglobe-utils

Closes #363 

Requires https://github.com/brainglobe/brainglobe-utils/pull/27 to be released and merged first

PR #366, opened 2024-01-16
Add curation keyboard shortcuts

### LINKED ISSUE #364: [Feature] Add keyboard shortcuts to the curation interface

adamltyson:
Requires #365 to be merged first

Closes #364

### DIFF
--- a/cellfinder/napari/curation.py
+++ b/cellfinder/napari/curation.py
@@ -70,6 +70,7 @@ class CurationWidget(QWidget):
         self.output_directory: Optional[Path] = None
 
         self.setup_main_layout()
+        self.setup_keybindings()
 
         @self.viewer.layers.events.connect
         def update_layer_list(v: napari.viewer.Viewer):
@@ -182,6 +183,7 @@ class CurationWidget(QWidget):
             self.load_data_layout,
             self.mark_as_cell,
             row=5,
+            tooltip="Mark all selected points as non cell. Shortcut: 'c'",
         )
         self.mark_as_non_cell_button = add_button(
             "Mark as non cell(s)",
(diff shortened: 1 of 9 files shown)


cellfinder/napari/curation.py
```diff
@@ -373,16 +384,45 @@ def save_training_data(
...
             self.update_status_label("Ready")
 
+    def __prep_directories_for_save(self):
+        self.yaml_filename = self.output_directory / "training.yml"
+        self.cell_cube_dir = self.output_directory / "cells"
+        self.no_cell_cube_dir = self.output_directory / "non_cells"
+
+        self.__delete_existing_saved_training_data()
+
+    def __delete_existing_saved_training_data(self):
```
alessandrofelder, line 415: For my understanding, why do functions in the file start with two underscores??

alessandrofelder (approved):
Looks good!

Just making a note here for our future selves and other interested people, that this PR mainly
* adds keyboard shortcuts
but also
* replaces local functions with `brainglobe-utils` equivalents
* removes tests for those functions
* makes tiny improvements to explicitness of code
* adds a "overwrite existing data?" warning.

cellfinder/napari/curation.py
adamltyson: Double underscores [are as close to private as you can get in Python](https://docs.python.org/3/tutorial/classes.html#private-variables). Not sure why they're used here and not anywhere else in BrainGlobe though.

alessandrofelder: Ah cool! Didn't know. From reading that it sounds like the use case is mainly to make it impossible to override this in subclasses. We currently don't have any subclasses of `CurationWidget`, and for consistency it would be good to not use it here, but then again not sure it's enough of a priority right now :thinking: 

TLDR I suggest leaving for now?

adamltyson: Fine by me.

adamltyson:
The tests are failing now, but is this due to #368?

adamltyson:
Yeah there's no rush for this

PR #369, opened 2024-02-01
Remove `pytest-lazy-fixture` due to incompatibility with `pytest v8.0.0`

### LINKED ISSUE #368: [BUG] CI breaks with pytest v8.0.0
CI (specifically pytest) now fails in the setup phase since version 8 was released. This is likely due to the changes to the collection phase, although I haven't had time to delve into the details.

Regardless, the tests themselves are not failing - rather pytest is failing to setup the tests themselves.

Failing workflow: https://github.com/brainglobe/cellfinder/actions/runs/7740974117/job/21107107180

Pytest version 8 release notes: https://docs.pytest.org/en/latest/changelog.html

willGraham01:
Resolves #368 |

Turns out we were only using `pytest-lazy-fixture` for one test anyway.

Workaround is to pass the fixture name to `pytest.param` as a string, and then use the `requests` fixture (built in to `pytest`) to fetch the value of that fixture.

As such, `pytest-lazy-fixture` is no longer a dependency.

Also fixes:

- We're not shipping `scikit-image` even though cellfinder depends on it?


alessandrofelder:
Looks like tests on Ubuntu for Python 3.10 reproducibly encounter a segmentation fault :thinking: would you be able to investigate further, @willGraham01 , please?

willGraham01:
> Looks like tests on Ubuntu for Python 3.10 reproducibly encounter a segmentation fault 🤔 would you be able to investigate further, @willGraham01 , please?

Will do this afternoon / Monday after DocSkills teardown admin :+1:

pyproject.toml
```diff
@@ -46,11 +46,11 @@ dev = [
     "pre-commit",
     "pyinstrument",
     "pytest-cov",
-    "pytest-lazy-fixture",
     "pytest-mock",
     "pytest-qt",
     "pytest-timeout",
     "pytest",
+    "scikit-image",
```
sfmig, line 53: I think this `scikit-image` can be removed from this dev list - it is in the dependencies list above as well

willGraham01: Gracias, me culpa for merging another fix into `main` earlier :sweat_smile:

willGraham01:
Tests seem to be passing now, but leaving this PR open due to the presence of #372 and it's close ties to that.

But if people need CI to pass, consider merging or rebasing onto this branch for the time being.

willGraham01:
Investigations over on https://github.com/brainglobe/cellfinder/pull/376 seem to imply that the garbage collection fix was not necessary (all tests seem to be passing without the need for the garbage collection cleanup statement).

I suspect there might be a faulty cache, so have deleted the cache for this branch and requested a re-run of the original CI that threw the seg-fault error: https://github.com/brainglobe/cellfinder/actions/runs/7743237973/job/21314877990

alessandrofelder (approved):

alessandrofelder:
Is the 20 min timout limit still sensible here, @willGraham01 ?

alessandrofelder:
Once this is merged, we should run the CI again for #366 #367 and #370 (and possibly others). If the tests pass #366 and #370 can be merged (as they are already approved).

PR #370, opened 2024-02-02
Update test_and_deploy.yml

alessandrofelder:
See https://github.com/neuroinformatics-unit/movement/pull/108


IgorTatarnikov (approved):
This looks good, is it worth waiting on the tests to be fixed, or should we just merge this?

PR #373, opened 2024-02-06
Smg/keras3 tf backend

sfmig:
Before submitting a pull request (PR), please read the [contributing guide](https://github.com/brainglobe/.github/blob/main/CONTRIBUTING.md).

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Addition of a new feature
- [ ] Other

**Why is this PR needed?**

**What does this PR do?**

## References

Please reference any existing issues/PRs that relate to this PR.

## How has this PR been tested?

Please explain how any new code has been tested, and how you have ensured that no existing functionality has changed.

## Is this a breaking change?

If this PR breaks any existing functionality, please explain how and why.

## Does this PR require an update to the documentation?

If any features have changed, or have been added. Please explain how the documentation has been updated (and link to the associated PR). See [here](https://brainglobe.info/community/developers/index.html#to-improve-the-documentation) for details.

## Checklist:

- [ ] The code has been tested locally
- [ ] Tests have been added to cover all new functionality (unit & integration)
- [ ] The documentation has been updated to reflect any changes
- [ ] The code has been formatted with [pre-commit](https://pre-commit.com/)

PR #374, opened 2024-02-06
Test Keras is present

sfmig:
This PR:
- Updates the dependencies and dev dependencies to work with Keras 3.0
- Replaces the Tensorflow check with a Keras check (tested in CI).

**Notes:**
- To avoid TF [overwriting](https://keras.io/getting_started/#installing-keras-3) the Keras version to 2.x, I pinned the TF version to a specific 2.16-nightly for now (same as they do in the [keras repo](https://github.com/keras-team/keras/blob/master/requirements.txt)).
- Before migrating to Keras 3.0, we checked that cellfinder tools throw an error if Tensorflow is not present. This PR replaces this check with an equivalent check for Keras.
- The CI checks that are failing (running package tests, brainmapper tests and running tests without numba) are dealt with in a [follow-up PR](https://github.com/brainglobe/cellfinder/pull/373).

**Question for reviewer**: 
Keras still needs a backend to work (for now tensorflow) - should we also check that one backend is defined? More on setting the backend [here](https://keras.io/getting_started/#configuring-your-backend).

### DIFF
--- a/.github/workflows/test_include_guard.yaml
+++ b/.github/workflows/test_include_guard.yaml
@@ -35,8 +35,8 @@ jobs:
             import cellfinder.core
             import cellfinder.napari
 
-      - name: Uninstall tensorflow
-        run: python -m pip uninstall -y tensorflow
+      - name: Uninstall keras
+        run: python -m pip uninstall -y keras
 
       - name: Test (broken) import
         id: broken_import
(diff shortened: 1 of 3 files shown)


IgorTatarnikov (approved):
Looks great! 👍 . I think we can leave out the backend check for now, especially if the goal is to make it backend agnostic in the end.

Issue #372, opened 2024-02-06
[BUG] Threading Segmentation Faults


willGraham01:
First encountered [here](https://github.com/brainglobe/cellfinder/pull/369#issuecomment-1923605577) after attempting a fix for #368 |

From what we can infer; it appears that some combination of our multiprocessing, CPU allocation/requesting, and Python's garbage collection process is resulting in segmentation faults during _teardown_.

This only occurs on Ubuntu machines running Python 3.10 specifically, all other OSes are fine, as are other Python versions. In particular though, Python 3.10 is what BrainGlobe users on the HPC system have access too - and @alessandrofelder has received similar bug reports from users.

## The symptoms

Identifiable symptoms seem to be: after running the detection algorithm, but before completing teardown, Python will hang due to a segmentation fault in the (memory) cleanup phase.

This is most easily replicated by running the `tox` command to execute the tests on a GitHub runner. Note that the segmentation fault message will appear some time _after_ the first `test_detection` test runs.

- The fix on https://github.com/brainglobe/cellfinder/pull/369 seems to be to explicitly invoke the garbage collector at the end of the detection algorithm.
- Further evidence from the SLURM job outputs received by @alessandrofelder also suggest that the detection algorithm runs to completion, but then hangs at teardown due to bad/missing memory references.

## Minimum failing example

Will be pasted here once a script that reliably reproduces the error is available.

## Possible causes

- Multiprocessing memory management
- Delegation of tasks to cores
- Use of the `cpu_count` method allocating CPUs, and claiming they are "free", before garbage collection can be run

## Ruled out causes

- We are NOT requesting more CPUs than we have available, so this is not the cause of the problem
- The detection algorithm itself manages to complete without error, the issue always occurs after it finishes.

alessandrofelder:
(Assuming what I observe on our local cluster is the same bug) and running with the latest changes from #375, it looks like https://github.com/brainglobe/cellfinder/blob/68ccb12d128c49eba75c1c916c36d6f8aa7e7d60/cellfinder/core/detect/filters/volume/volume_filter.py#L147 is reached, but https://github.com/brainglobe/cellfinder/blob/68ccb12d128c49eba75c1c916c36d6f8aa7e7d60/cellfinder/core/detect/filters/volume/volume_filter.py#L198 is not...

This is part of the code that happens still in the multiprocessing context, but should[ just happen in the main process.](https://brainglobe.info/community/developers/repositories/cellfinder-core/index.html#id1)

[...]

PR #376, opened 2024-02-07
Searching for min. segfault example

willGraham01:
# DO NOT MERGE!

At present, #372 seems to be only replicable on GH runners, so creating this branch with a more focused CI workflow to attempt to distil a minimum failing example.


willGraham01:
[CI passes on this commit](https://github.com/brainglobe/cellfinder/pull/376/commits/f49ecfb35cf71ccff12afa1139adac0635788eb2), which should be functionally identical to the faulty branch without the fix. Further investigation is being conducted.

PR #379, opened 2024-02-08
Cellfinder with Keras 3.0 and jax backend

sfmig:
With this PR, cellfinder tests pass with Keras 3.0 and both the tensorflow and the JAX backend 🥳 

Specifically:
- I added checks similar to previously existing ones, that checked if `tensorflow` was installed in the environment before importing `cellfinder`. Now we check if `keras` is installed, and if it has an appropriate backend installed. I modified the CI check accordingly.

- I configured tox to run tests using the tensorflow backend and the JAX backend.

- for the brainmapper tests we run in CI, I added a `pip install` step that installs the (for now pinned) tensorflow version compatible with Keras 3.0.0. I am (temporarily) setting tensorflow as the default backend for Keras if none is defined.

- for the time being I pinned Keras and TF to specific versions, but happy to inspect that further in a future PR (in theory TF 2.15 [should be compatible](https://keras.io/getting_started/#compatibility-matrix) with Keras 3.0.0)

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -115,8 +115,10 @@ jobs:
       - name: Install test dependencies
         run: |
           python -m pip install --upgrade pip wheel
-          # Install latest SHA on this brainglobe-workflows branch
+          # Install cellfinder from the latest SHA on this branch
           python -m pip install git+$GITHUB_SERVER_URL/$GITHUB_REPOSITORY@$GITHUB_SHA
+          # Install tensorflow as keras' default backend
+          python -m pip install "tf-nightly==2.16.0.dev20240101"
           # Install checked out copy of brainglobe-workflows
           python -m pip install .[dev]
 
--- a/.github/workflows/test_include_guard.yaml
+++ b/.github/workflows/test_include_guard.yaml
@@ -24,7 +24,7 @@ jobs:
         with:
           python-version: '3.10'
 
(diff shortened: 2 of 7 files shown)


IgorTatarnikov (approved):
👍 Looks great! Just tried it and cell detection seems to work with a jax backend using the pretrained model.

Just a few nit picky comments but nothing that has to be changed.

adamltyson:
> Just tried it and cell detection seems to work with a jax backend using the pretrained model.

To clarify, this is the existing pre-trained model, but interference being run with JAX not TF? Awesome! I assumed this wouldn't be possible. Does it need retraining for torch?

IgorTatarnikov:
I was quite surprised as well. The only change I had to make was to set the `KERAS_BACKEND` environmental variable to jax in a fresh environment and I was able to do inference with the default cellfinder settings!

I'm not sure about torch, @sfmig mentioned she was more issues trying to get cellfinder to run with a torch backend but I'm not sure what the status of that is.

sfmig:
> I'm not sure about torch, @sfmig mentioned she had some issues trying to get cellfinder to run with a torch backend but I'm not sure what the status of that is.

yes, torch wasn't as smooth as JAX on a first try so started with the easy one 😇 

> To clarify, this is the existing pre-trained model, but interference being run with JAX not TF?

yes, it is the pre-trained model but we are passing `test_train` which I think is finetuning the pre-trained model to new data right?

PR #394, opened 2024-03-26
basic fix avoiding writing to package files

### LINKED ISSUE #393: [BUG] Writing custom config is not possible on HPC with shared installation of BrainGlobe
**Describe the bug**
The default paths returned in https://github.com/brainglobe/cellfinder/blob/main/cellfinder/core/tools/source_files.py are not (necessarily) writeable by users (for good reason). It's generally a bad idea for users to write into the package files. The default should presumably point somewhere else. 

**To Reproduce**
Run `brainmapper` tool on HPC as user (with no write permissions to BrainGlobe module!)

**Expected behaviour**
It is possible for users to write custom cellfinder configurations.

alessandrofelder:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
Specifying a non-default model currently writes to package files (which is a bad idea in general, but in particular causes errors on shared systems like our local HPC).

**What does this PR do?**
User-specified model configuration files are written to user folder instead.
It also documents and renames some of the functions involved.

## References

Closes #393

Existing tests pass.

## Is this a breaking change?
No.

No.

## Checklist:

### DIFF
--- a/cellfinder/core/tools/source_files.py
+++ b/cellfinder/core/tools/source_files.py
@@ -11,12 +11,12 @@ def default_configuration_path():
     return Path(__file__).parent.parent / "config" / "cellfinder.conf"
 
 
-def custom_configuration_path():
+def user_specific_configuration_path():
     """
-    Returns the path to the custom configuration file for cellfinder.
+    Returns the path to the user-specific configuration file for cellfinder.
 
-    This function returns the path to the custom configuration file
-    for cellfinder. The custom configuration file is located in the
+    This function returns the path to the user-specific configuration file
+    for cellfinder. The user-specific configuration file is located in the
     user's home directory under the ".cellfinder" folder and is named
     "cellfinder.conf.custom".
 
--- a/cellfinder/core/tools/source_files.py
+++ b/cellfinder/core/tools/source_files.py
@@ -6,4 +6,4 @@ def source_config_cellfinder():
 
 
 def source_custom_config_cellfinder():
-    return Path(__file__).parent.parent / "config" / "cellfinder.conf.custom"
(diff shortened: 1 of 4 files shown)


alessandrofelder:
I am happy to adapt brainmapper to the new names and re-release once/if my suggested variable/function names in this PR have been approved by a reviewer, and make the related releases.

cellfinder/core/tools/source_files.py
```diff
@@ -1,9 +1,27 @@
...
+    This function returns the path to the user-specific configuration file
+    for cellfinder. The user-specific configuration file is located in the
+    user's home directory under the ".cellfinder" folder and is named
+    "cellfinder.conf.custom".
+
+    Returns:
+        Path: The path to the custom configuration file.
+
+    """
+    return Path.home() / ".cellfinder" / "cellfinder.conf.custom"
```
willGraham01, line 27: I seem to recall we have an outstanding issue to standardise where we put user configs; and want to say we ended up settling on the `~/.brainglobe` directory. So maybe this should be `~/.brainglobe/cellfinder`? Though I don't know what the current default directory for `brainmapper` is set to.

willGraham01 (approved):
I can't find the relevant issue or PR, but I have a feeling that we switched from using `~/.cellfinder` to `~/.brainglobe` at some point.

cellfinder/core/tools/source_files.py
alessandrofelder: Good point... (relevant issue is https://github.com/brainglobe/BrainGlobe/issues/26) - I suggest tackling that in one go, and for now sticking to the "old" convention?

Issue #404, opened 2024-04-18
[BUG] Intermittent Segmentation Faults on CI


willGraham01:
Moving the discussion from https://github.com/brainglobe/cellfinder/pull/403 here.

[First reported](https://github.com/brainglobe/cellfinder/pull/403#issuecomment-2059253491): @K-Meech

> That seems to fix codecov! No more codecov complaints on the latest actions run. The run tests with numba disabled action is hanging though, [due to a segmentation fault](https://github.com/brainglobe/cellfinder/actions/runs/8705982093/job/23881108160?pr=403#step:5:224). Not sure what is causing this - it seems unrelated to any changes in this PR.

@IgorTatarnikov 

> The run tests with numba disabled segfault seems to be sporadic. I ran into the same issue [here](https://github.com/brainglobe/cellfinder/actions/runs/8693597959/attempts/2) but re-running it made it happy. Not sure what's going on!

@willGraham01 

> My only guess off the top of my head would be that the runs got runners with different specs (maybe an older runner vs a newer one)? But looking at the attempts Igor linked to, it seems it's the same machine (specs-wise) so that rules that out.
>
> Can we replicate this seg-fault locally? Or is it just a CI thing?

@IgorTatarnikov 

> ... Seems I marked my previous as duplicate and can't undo it. Reposting for posterity's sake.
>
> I tried running the tests locally on Ubuntu 22.04 having set NUMBA_DISABLE_JIT=1 but I couldn't replicate the seg fault over 5 runs. I first ran all tests, then focused on running just test_detection.py to save time. Seems it might be a CI thing?

@willGraham01

> A brief glance [at the logs](https://github.com/brainglobe/cellfinder/actions/runs/8693597959/job/23875579407#step:5:246) indicates that the seg-fault is thrown during garbage collection whilst in a multiprocessing thread.
>
> Python's garbage collector is not deterministic, which is the main motivator behind my following guesses 😅
>
> - One thread is finishing / exiting earlier than the others, and cleans up some (implicitly) shared memory before the other threads can finish? This is also supported by Kimberly's discovery that the most recent test hung for 6hrs then was killed - we might end be ending up in a deadlock caused by something like this.
> - We're not respecting the private/shared memory for each thread. Don't know if Python multiprocessing cares about these concepts in the same way that something like C++ does, though.
> - Some combination of the above combined with how GitHub runners handle threading.
>
> But I could be well off the money in each of those though.

willGraham01:
[...]

PR #406, opened 2024-04-23
Remove modular asv benchmarks

sfmig:
keeping an eye on brainglobe/cellfinder#404 👀 

## Description

**What is this PR**

- [ ] Bug fix
- [x] Other

**Why is this PR needed?**
To remove unused benchmarking code.

**What does this PR do?**
Deletes old code that used `asv` to benchmark individual functions of `cellfinder.core`. 

We decided to go for an integrated approach to benchmarking, in which we benchmark a full workflow rather than modules or individual functions. Because of this, we also moved all benchmarking code to [the workflows repository](https://github.com/brainglobe/brainglobe-workflows).

## References

The most relevant issues are brainglobe/brainglobe-workflows#105 and brainglobe/brainglobe-workflows#106, although since then the benchmarking work has moved to [the workflows repository](https://github.com/brainglobe/brainglobe-workflows).

The code has been tested locally.

## Is this a breaking change?

No.

No.

## Checklist:


alessandrofelder (approved):

alessandrofelder:
PRs that remove things are always nice 🤩 
I've transferred the issues over to the workflows repo... feel free to close them as duplicates if you already have equivalent issues there, @sfmig !

adamltyson:
> We decided to go for an integrated approach to benchmarking, in which we benchmark a full workflow rather than modules or individual functions. Because of this, we also moved all benchmarking code to [the workflows repository](https://github.com/brainglobe/brainglobe-workflows).

Late to the party here, but does this mean that if some changes are made here, we won't find out if they affect the benchmarks until they are run on the workflows repo? I.e. rather than being part of the CI?

willGraham01:
We have [a check in CI](https://github.com/brainglobe/cellfinder/blob/6b529dc72110114f0d17c8139e8645f8fecef317/.github/workflows/test_and_deploy.yml#L101) to make sure that `brainmapper` isn't breaking. We could generalise it to other benchmarks if we want to.

adamltyson:
That would be great if it's not super complicated. The main reason for this effort is to catch performance degradations early.

PR #408, opened 2024-04-29
Adapt CI so it covers both new and old Macs, and installs required additional dependencies on M1

alessandrofelder:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

`macos-latest` is now a Silicon Mac runner, so we need to adapt CI accordingly

**What does this PR do?**
* ensures CI is run on both Mac chip architectures
* runs tensorflow include guard tests on Silicon mac instead.

## References

Partial solution to https://github.com/brainglobe/BrainGlobe/issues/63

CI that was breaking before, now passes

## Is this a breaking change?

Now

No

## Checklist:

- [n/a] The code has been tested locally
- [n/a] Tests have been added to cover all new functionality (unit & integration)
- [n/a] The documentation has been updated to reflect any changes

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -42,12 +42,14 @@ jobs:
         # Run all supported Python versions on linux
         os: [ubuntu-latest]
         python-version: ["3.9", "3.10"]
-        # Include one windows, one macos run
+        # Include one windows, one macos run each for M1 (latest) and Intel (13)
         include:
-        - os: macos-latest
-          python-version: "3.10"
-        - os: windows-latest
-          python-version: "3.10"
+          - os: macos-13
+            python-version: "3.10"
+          - os: macos-latest
+            python-version: "3.10"
+          - os: windows-latest
+            python-version: "3.10"
 
     steps:
       # Cache the tensorflow model so we don't have to remake it every time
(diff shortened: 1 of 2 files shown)


.github/workflows/test_and_deploy.yml
```diff
@@ -42,12 +42,14 @@ jobs:
...
-        - os: macos-latest
-          python-version: "3.10"
-        - os: windows-latest
-          python-version: "3.10"
+          - os: macos-13
+            python-version: "3.10"
+          - os: macos-latest
+            python-version: "3.10"
+          - os: windows-latest
+            python-version: "3.10"
```
sfmig, line 52: question: do you know why we don't need to install `hdf5` here?

sfmig (approved):
thanks @alessandrofelder, looks good too! 😁 Just a couple of questions out of curiosity.

.github/workflows/test_and_deploy.yml
alessandrofelder: The `hdf5` installation issue on M1 is linked to depending on `pandas`, which cellfinder doesn't.

.github/workflows/test_include_guard.yaml
```diff
@@ -35,7 +35,12 @@ jobs:
             import cellfinder.core
             import cellfinder.napari
 
-      - name: Uninstall tensorflow
+      - name: Uninstall tensorflow-macos on Mac M1
+        if: matrix.os == 'macos-latest'
```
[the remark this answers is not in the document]
alessandrofelder: We install a [different `tensorflow` depending on ARM and Intel architectures](https://github.com/brainglobe/cellfinder/blob/0fd0a5e3712047b5cac138f329a3c70dd48bca66/pyproject.toml#L33), and I wanted to cover both with the same number of runs as before?

PR #414, opened 2024-05-07
Simplify model download

### LINKED ISSUE #268: Save models into e.g. `.brainglobe/cellfinder/models`
See https://github.com/brainglobe/BrainGlobe/issues/26

### LINKED ISSUE #296: Verify download of pre-trained model
If the download of the pre-trained model is interrupted, it can't be loaded by tensorflow and there's a fairly cryptic error message (e.g. `OSError: Unable to open file (truncated file: eof = 62652416, sblock->base_addr = 0...`) that doesn't help the user debug.

We could checksum, or even just catch the error. Catching the error could be used for any model/weights, not just the pre-trained model.

### LINKED ISSUE #390: Downloading the trained model crashes if the model directory already exists
e.g. if the previous download was corrupted, and deleted by the user. 

cc @IgorTatarnikov

adamltyson:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other

**Why is this PR needed?**
The code to download trained models was unnecessarily complicated and had some bugs. This was leftover from when the same code also downloaded atlases (pre `brainglobe-atlasapi`). 

**What does this PR do?**
* Uses Pooch to download the models
* Refactors and simplifies some of the other downloading code
* Saves models into `.brainglobe/cellfinder/models` by default
* Adds checksum for the default model
* Adds a download progress bar

## References
Closes #268
Closes #296
Closes #390

Automated tests have been updated and pass. Manual testing ongoing (it's very slow to properly test all scenarios as it involves downloading data). 

## Is this a breaking change?

No

No, but the `cellfinder_download` CLI should probably be documented. I've raised https://github.com/brainglobe/brainglobe.github.io/issues/176

## Checklist:


alessandrofelder (approved):
I think this is a useful simplication of the code, and looks good to me.
I briefly considered whether we should be worried about backwards-compatibility, but I don't think this will be a problem (there may just be a `.cellfinder/` and `.brainglobe/` folder on some machines for a while 🤷 )

Issue #266, opened 2023-11-29
Update to Python 3.11


adamltyson:
brainglobe/cellfinder#293 means that [we can't support Python 3.11](https://github.com/brainglobe/cellfinder-core/pull/228#issuecomment-1832285017) as wheels aren't available for old versions of TensorFlow.

PR #418, opened 2024-05-10
Updating to Keras 3.0 and migrating to PyTorch

### LINKED ISSUE #279: Consider moving away from TensorFlow
TF is causing us some headaches, namely lack of support for Windows, and not being on conda-forge. There are other options, but PyTorch seems to be the leading candidate currently. It's on conda-forge, seems fairly easy to install, and it's now part of the Linux Foundation (so can't be killed off by Meta). 

I don't know how easy the TF->PyTorch transition would be, but the 3D ResNet here isn't that complicated. A re-write may be a way of speeding things up. 

I remember someone had started looking into this (and reported some speedups), but I can't remember who!

IgorTatarnikov:

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
`tensorflow` has become increasingly difficult to support (e.g. lack of GPU support on native Windows). Switching to `keras` 3.0 allows us to migrate to using `torch` as the backend instead of `tensorflow`. This will make future maintenance easy and allow us to support Python 3.11+.

**What does this PR do?**
Upgrades to `keras` 3.0.
Sets `torch` as the default backend.
Removes any functions related to `tensorflow` (error suppression etc...)

## References
Closes #279
https://github.com/brainglobe/brainglobe.github.io/issues/177
https://github.com/brainglobe/brainglobe.github.io/issues/183
#266 

All tests pass on CI, basic workflows have been tested by manual inspection.

## Is this a breaking change?
No.

Yes, see https://github.com/brainglobe/brainglobe.github.io/issues/183

## Checklist:


IgorTatarnikov:
To do:
Documentation:
- [ ] Update documentation on the [brainglobe website](https://brainglobe.info), tracked in brainglobe/brainglobe.github.io/issues/183
- [ ] Prep blog post? Not sure if this needs to be done in sync with the merge into main, or the release of v1.2.0

Sanity checks for a "regular" cellfinder workflow (should be done via the `napari` GUI or via the `brainmapper` CLI where possible):
- [ ] Detect using the default `cellfinder` model
- [ ] Detect using a custom model
- [ ] Curate a set of cells and use it to retrain the default model
- [ ] Detect using the new updated model

IgorTatarnikov:
Tests are currently not running `mac-latest` since I was having issues getting `torch` to behave on CI, see [here](https://github.com/brainglobe/cellfinder/actions/runs/9016187794/job/24772298500). I couldn't find an elegant way of forcing torch to not use the `mps` device on CI specifically while still allowing it to be used normally.

Not sure how to proceed there. Tests pass locally when run on my personal machine (M2 MacBook Pro running macOS 14.4.1).

adamltyson:
> Tests are currently not running mac-latest since I was having issues getting torch to behave on CI, see [here](https://github.com/brainglobe/cellfinder/actions/runs/9016187794/job/24772298500). I couldn't find an elegant way of forcing torch to not use the mps device on CI specifically while still allowing it to be used normally.
> Not sure how to proceed there. Tests pass locally when run on my personal machine (M2 MacBook Pro running macOS 14.4.1).

Naively, it seems like it should be possible (we're not the only people using torch & GH actions!). Ofc you've looked into it, so it's not simple.

I would suggest:
1. Give it a bit more of a go to see if you can fix it (there's a large online torch community)
2. If 1. fails, write up an issue with what you've tried and don't let this PR get derailed by a relatively small problem

cellfinder/__init__.py
```diff
@@ -1,25 +1,50 @@
...
+        f"For more information on brainglobe, please see "
+        f"https://github.com/brainglobe/brainglobe-meta#readme."
+    ) from e
+
+
+# If no backend is configured and installed for Keras, tools cannot be used
+# Check if backend is configured. If not, set to "torch"
+if not os.getenv("KERAS_BACKEND"):
+    os.environ["KERAS_BACKEND"] = "torch"
+    warnings.warn("Keras backend not configured, automatically set to torch")
```
adamltyson, line 32: I wouldn't have this warning by default. Every user will see this warning, but they won't actually need to do anything about it.

tests/core/test_integration/test_train.py
```diff
@@ -35,5 +35,5 @@ def test_train(tmpdir):
     sys.argv = train_args
     train_run()
 
-    model_file = os.path.join(tmpdir, "model.h5")
+    model_file = os.path.join(tmpdir, "model.keras")
```
adamltyson, line 38: is this the new default extension? This should be added to https://github.com/brainglobe/brainglobe.github.io/pull/189 (I think there's a couple of places in the docs that reference the .h5 files directly).

adamltyson:
@IgorTatarnikov I compared this PR with version 1.2.0 and the classification looks considerably worse. This was using the napari plugin, all parameters default, using the pre-trained model. 

The only other difference was the Python version (old cellfinder env is 3.10, and the torch version is 3.12), but that shouldn't affect this.

Do you know what's going on?

TensorFlow
tensorflow

Torch
torch

cellfinder/__init__.py
IgorTatarnikov: Done

adamltyson:
The classification problem seems to be resolved on that machine by deleting and re-downloading the model, not sure what's going on.

> Also, if I try to run pytest locally, I get a huge stacktrace that leads to a keras error: ModuleNotFoundError: No module named 'tensorflow'

This issue remains, the backend needs to be set explictly by editing the `~.keras/keras.json` file (I needed to do this on two machines). Can we override this?

adamltyson:
On my mac (m1), I get this error:

> E       NotImplementedError: Exception encountered when calling MaxPooling3D.call().
E       
E       The operator 'aten::max_pool3d_with_indices' is not currently implemented for the MPS device. If you want this op to be added in priority during the prototype phase of this feature, please comment on https://github.com/pytorch/pytorch/issues/77764. As a temporary fix, you can set the environment variable `PYTORCH_ENABLE_MPS_FALLBACK=1` to use the CPU as a fallback for this op. WARNING: this will be slower than running natively on MPS.

I can fix it for detection by setting `PYTORCH_ENABLE_MPS_FALLBACK=1`, however this isn't ideal because a) others will just see a big red error, and b) presumably we want to use MPS?

Also, even then, upon training, I get this error:

> TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.

adamltyson:
After some more testing, everything seems to work fine on my mac, the only issue is automatically setting the backend to torch.

tests/core/test_integration/test_train.py
IgorTatarnikov: If we're saving the whole model then we have to use the `.keras` extension. For weights, the files must now end with `.weights.h5`. I'll double check for references to `.h5` in our documentation.

adamltyson (approved):
This is great, thanks @IgorTatarnikov @sfmig!

I've tested it on macOS, Ubuntu & Windows now. It works fine across all OSs, but it does seem to be very slow on Windows. I think this is out of scope for this review, so I've raised https://github.com/brainglobe/cellfinder/issues/426 so we can look into it later on.

I think the only remaining thing is setting the backend automatically. If this can be sorted, we can create an rc and get some more people to test upgrading.

IgorTatarnikov:
I'm not sure why the automatic backend setting isn't working as it stands. I'll remove the conditional and always set the `KERAS_BACKEND` variable to be `torch` when the package is imported. This removes some flexibility but we weren't officially planning to support custom backends so if someone needs it they may have to edit the package code locally, or fork it.

adamltyson:
Is it worth getting rid of the other code in the init file now? There's not a lot of point setting the backend to torch, then checking whether it's TF or JAX.

IgorTatarnikov:
I think it's ready! Once this is merged we can merge [this](https://github.com/brainglobe/brainglobe-workflows/pull/112) PR in `brainglobe-workflows` to fix the brainmapper tests

PR #422, opened 2024-05-20
Update caching

adamltyson:
Copied from https://github.com/brainglobe/brainglobe-atlasapi/pull/301, now everything is in .brainglobe.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -52,12 +52,13 @@ jobs:
             python-version: "3.10"
 
     steps:
-      # Cache the tensorflow model so we don't have to remake it every time
-      - name: Cache tensorflow model
+      - name: Cache atlases
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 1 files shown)


IgorTatarnikov (approved):
Looks good :+1:

.github/workflows/test_and_deploy.yml
```diff
@@ -52,12 +52,13 @@ jobs:
             python-version: "3.10"
 
     steps:
-      # Cache the tensorflow model so we don't have to remake it every time
-      - name: Cache tensorflow model
+      - name: Cache atlases
```
IgorTatarnikov, line 55: Should this be changed to something more descriptive? I don't think we're actually using any atlases in the `cellfinder` tests so it would be more appropriate to call the step `Cache models` or even `Cache .brainglobe directory`

adamltyson: Good idea, updated.

alessandrofelder (approved):

PR #427, opened 2024-05-28
Update __init__.py

adamltyson:
The list of specific authors is misleading, as many more people have contributed greatly to this package. I propose the list here because it can never be representative.


alessandrofelder (approved):
I am confused by the test failures (connectivity issues or unrelated bug) but I am sure they are unrelated to these changes.

Issue #284, opened 2023-04-28
De-duplicate test dependency lists


dstansby:
Dependencies for testing are listed in two places:

https://github.com/brainglobe/cellfinder-core/blob/33f61dd98e44b55fc2017c33433b7fc150259880/pyproject.toml#L44-L53

https://github.com/brainglobe/cellfinder-core/blob/33f61dd98e44b55fc2017c33433b7fc150259880/tox.ini#L14-L17

Ideally these should be de-duplicated

PR #431, opened 2024-05-31
Remove test dependency duplication

adamltyson:
We shouldn't need to define test dependencies in tox. This PR relies on installing the `dev` and `napari` extras, so hopefully we have one less thing to keep up to date.

Closes https://github.com/brainglobe/cellfinder/issues/284


IgorTatarnikov (approved):
Looks good :+1:

Issue #353, opened 2020-02-27
[Feature] Need an intelligent way of picking batch size.


adamltyson:
The default batch size for classification (32) was chosen so it would run on pretty much any good GPU but as @larsrollik has found, this can be increased a lot, reducing interference time (128 works with the default resnet on a RTX 2080Ti).

Maybe use tf to get available GPU memory and pick batch size based on that? Allow users to specify max GPU memory to use?

IgorTatarnikov:
I haven't played with batch size during inference. In theory there should be a way to pre-calculate a maximal batch size to utilize as much VRAM as possible during inference. I'd want to do some profiling first though to see how much effect it has before going through the effort of implementing!

PR #432, opened 2024-05-31
Update default inference batch size

adamltyson:
This PR:
- Updates the default inference batch size to 64. This is only based on my own local experiments that this seems to have the best compromise of speed and RAM/VRAM usage. It seems to speed classification up by approx 40% (depending on the number of points per plane).
- Exposes the batch size parameter in the napari GUI so users can change this if needed.
- Prints timing of classification (as with detection) to help assess performance.

I've left the training batch size alone because training requires more memory, and the batch size actually affects training (unlike inference). 

Closes https://github.com/brainglobe/cellfinder/issues/353


IgorTatarnikov (approved):
Looks good, I tested this on Ubuntu with an RTX 4080, a batch size of 64 was ~50% faster for inference when compared to 16.

PR #458, opened 2024-09-20
Schedule weekly tests

IgorTatarnikov:
Schedule tests to run weekly to make sure there are no installation issues, and to keep the cache fresh.


IgorTatarnikov:
I've opened a PR in `brainglobe-workflows` to fix the test failure https://github.com/brainglobe/brainglobe-workflows/pull/128

PR #462, opened 2024-10-08
Pin `keras` version to 3.5.0

IgorTatarnikov:
Currently, training enters an infinite loop at the end of an epoch.

This only happens with `keras==3.6.0`. Temporarily pinning to `3.5.0`.


alessandrofelder (approved):
Awesome, thank you!

PR #486, opened 2025-02-26
Move qt-niu dependency

adamltyson:
I put it into the dev dependencies by mistake. Cellfinder should be re-released once this is merged. I've yanked `1.5.0` on PyPI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ dependencies = [
     "tifffile",
     "tqdm",
     "qt-niu"
-
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -34,6 +34,8 @@ dependencies = [
     "torch>=2.1.0,!=2.4",
     "tifffile",
     "tqdm",
+    "qt-niu"
+
```
IgorTatarnikov, line 38:
```suggestion
    "qt-niu"
```

IgorTatarnikov (approved):
Looks good, just a stray new line but that's not too important.

Issue #468, opened 2024-11-07
[Feature] Run cell detection on subvolume of an image and not just select z planes


IgorTatarnikov:
**Is your feature request related to a problem? Please describe.**
Currently, `cellfinder` only supports restricting the search space by the start and end plane. Sometimes, it would be nice to restrict the search space in plane as well (e.g., only run detection in (:, 100:400, 200:500) (z, y, x))

vijayalakshmipokala:
Hi @alessandrofelder, @adamltyson, @IgorTatarnikov, 

I’m a GSoC 2026 participant and I’d like to work on this issue as part of my BrainGlobe project on expanding input support in `cellfinder`.

I plan to implement:
- Optional `--x-range` and `--y-range` parameters to allow detection on subvolumes.
- Maintain backward compatibility with the existing `--z-range` functionality.
- Update the detection backend to only process the specified subvolume.
- Add automated tests verifying full-volume and subvolume detection.
- Update documentation and usage examples.

Please let me know if this approach looks good or if there are any additional requirements before I start a PR.

Thanks!

PR #491, opened 2025-03-08
[Feature] Run cell detection on subvolume of an image and not just select z planes #468

parharti:
This PR refactors the __get_oriented_image method to extract and rescale 3D image regions around a given cell without requiring explicit user-defined bounding boxes. The function now: #468 

- Dynamically determines bounding limits based on image_stack dimensions.
- Ensures the cropped region does not exceed image boundaries.
- Raises an error if the extracted region is empty.
- Supports augmentation if self.augment is enabled.
- Applies proper scaling and interpolation to maintain consistency.

### DIFF
--- a/cellfinder/core/classify/cube_generator.py
+++ b/cellfinder/core/classify/cube_generator.py
@@ -292,40 +292,37 @@ class CubeGeneratorFromFile(Sequence):
         )
         return images
 
-    def __get_oriented_image(
-        self, cell: Cell, image_stack: np.ndarray
-    ) -> np.ndarray:
-        # Define cropping region (Modify as needed)
-        z_min, z_max = 10, 100  # Crop along z-axis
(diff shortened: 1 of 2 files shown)


adamltyson:
Hi @parharti, I don't think this PR addresses #468. That issue about running the whole of cellfinder on a subset of an image loaded into the napari plugin. This PR seems change how the cubes are extracted for the later classification step.

parharti:
> Hi @parharti, I don't think this PR addresses #468. That issue about running the whole of cellfinder on a subset of an image (defined in all three axis). This PR seems change how the cubes are extracted for the later classification step.
ohh!!
i see
thank you for rewieng it 
i will get back with the solution after researching about it!!

parharti:
for now i think closing i PR would be a better option i think to keep issue section clean

PR #493, opened 2025-03-14
Optimize data loading for cell classification

matham:
## Description

This is still a WIP and needs a bit of polish.

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

When testing cell classification with our large whole brains (600GB+, 2.5 million+ potential cells), it took 32+ hours to run inference on all the potential cells. Looking at task manager, I could see that the GPU was barely utilized. This is typically an indication that the data was not read or processed quickly enough to keep the GPU fed.

With my changes, it only took `<1.5 hours`.

Also, when testing training the model with the original `serial2p` dataset, it took `79.6 min / epoch` for main, vs `24.33 min / epoch` with the changes. And it used less than half the RAM of main.

**What does this PR do?**

Very broadly, this PR refactors the data loading and augmentation classes so the various features can be shared among different datasets (dask arrays, tiff files containing cubes etc). But more importantly it adds caching to data loading. It also re-works the multi-processing of workers loading the data.

Consider a bunch of points with various `z` values. We need to get a cuboid centered on the point. Say in Z it's 20 planes. Then for every point we'd access the Dask array to get 20 planes and extract the cuboid. So for each point we potentially read 20 planes. We already sorted during inference to read sequential `z` points, but still unless Dask did some caching we'd be reading the same data constantly. And certainly there's overhead.

Instead, we cache the last `n` planes in memory and so when reading `z` sequentially we save a lot of disk/Dask access. This is also quite important when we split processing up among workers who may access neighboring `z` points, preventing Dask (OS) from caching data. But, having our cache twice the size of a cube makes this not a problem for us.

Additionally, there's a dedicated thread in the main process that does the data loading. All the worker (sub-processes) communicate with it via queues by sending it a buffer that the thread fills in with the loaded data and the worker than process it (e.g. augment, resize). pytorch tensors can be shared among processes in memory so this saves us a bunch of data copying.

--------

Important changes:

- The only new (potential) external parameter is the `max_workers`. Unlike elsewhere where we want to say how many cores we want to keep free, here it makes more sense to say at most how many workers we want to use. Because additional workers above like 6 (in my testing), was just wasted due to overhead (multiprocessing).
- Added `monai` as a dependency so we can use their random transforms during data augmentation. Their transforms are proper 3d.
- The data has 4 dimensions, channel, x, y, and z. Everyone has their own requirement for the order. Torch has its order, the data is saved on disk in its order, `monai` has its own order, and the model expected the data in yet another order. In main this order is hardcoded. But I added an explicitly defined ordering for each of them so we can automatically re-order the data as needed. It needs more code, but prevents mental complexity.
- When using arrays for the data, in main we filter out any points that are on the edge of the 3d stack. However it didn't take into account that we have to rescale the input data to the given network expected size. So this filtering was incorrect for point near the boundary when the voxel size of the data is different than the voxel size of the network (see the previous point). This has been fixed.
- For a given point, we find a cuboid centered around this point. In `main`. the way we center it is different in x, y vs z. I kept this difference so as to stay compatible with main as much as possible.

--------

Work to be done:

- Add docs
- Add tests
- Better compare cell classification for main vs this branch. Specifically check that using the existing weights classifies cells similarly for main and this branch.
- Compare a model trained with main vs this branch to see how similarly they classify the same cells.

## References

None

I tested it by training on the original `serial2p` dataset. You can see the train/test loss/accuracy to see it correctly trained:

train/val accuracy:
image

loss:
image

I would like to do more comparisons to see how points are (potentially) differently classified between main and this branch.

## Is this a breaking change?

The only potential breaking change is that the augmentation is different. But that shouldn't affect existing model inference, only newly trained models. At the same time, we need to double check this.

We may also want to consider retraining the model?

We potentially only need to add support for the `max_workers` parameter.

## Checklist:

- [ ] The code has been tested locally


matham:
I think one of the issues with the actions is https://github.com/Project-MONAI/MONAI/issues/8277. It lists an older numpy as a dependency even though it's compatible with the latest numpy.

adamltyson:
It's updated on main though, so hopefully the next release will work ok.

adamltyson:
Looks like the next release should be out soon: https://github.com/Project-MONAI/MONAI/issues/8421

alessandrofelder:
There seems to have been some movement on the MONAI repo last week, getting closer to a new release. Just commenting here to remind ourselves that this PR still exists :)

matham:
I'm hoping to start working on this again this week! 🤞

adamltyson:
The slow release cadence of monai has made me rethink having it as a dependency. @matham, do you have an idea of how complicated it would be to remove the dependency from this PR?

matham:
Originally, when I looked at it, I saw 3 options. The original scipy, `TorchIO`, and `monai`.

1. scipy didn't seem to do true 3d transformations - rather it treated a cube as a bunch of 2d images that is each transformed.
2. `TorchIO` used simpleitk, which I didn't love as it adds another large transitive dependency and its need to move data between torch to itk. Plus it had more dependencies overall: `deprecated, humanize, nibabel, numpy, packaging, rich, scipy, simpleitk, torch, tqdm, typer, wrapt, markdown-it-py, pygments, filelock, typing-extensions, networkx, jinja2, fsspec, sympy, mpmath, colorama, typer-slim, typer-cli, click, shellingham, mdurl, MarkupSafe`.
3. `monai` worked natively with torch without major dependencies: `numpy, torch, filelock, typing-extensions, networkx, jinja2, fsspec, sympy, mpmath, MarkupSafe` so I went with that.

I looked at extracting the relevant code from monai, which should be doable. But I didn't think you'd want to have to maintain that so I didn't do it. But perhaps you do prefer it? Ultimately for `monai` it comes down to building transforms and applying it. But my linear algebra is not amazing, especially in 3d. But I can give it a try?

There are other libraries that can handle 3d transformations, e.g. `pytransform3d`, or even scipy, but they transform vectors, not volumes. You can use that to [transform volumes](https://stackoverflow.com/questions/40524490/transforming-and-resampling-a-3d-volume-with-numpy-scipy) but it seems a bit out there and prone to issues.

adamltyson:
Hmm, lets go ahead with monai, and vendor the necessary code down the line if we feel the need to.

alessandrofelder:
(PS Also, if you could pick up the merge conflict... think it should be straightforward to resolve)

matham:
I finished adding full tests and docs to augment. So it should all now be good to go. There's 100% coverage on both augment and cube_generator.

matham:
A *seeming* difference with this PR: `random_bool` previously was flipped (before https://github.com/brainglobe/cellfinder/pull/581). So `0.1` augmentation probability actually resulted in `0.9` augmentation. The default augmentation value for training was `0.1`, meaning in reality, when the pre-trained model trained, it did `90%` augmentation. This is the condition under which the pre-trained model was trained.

In this PR, I didn't fix `random_bool`, instead I made the defaults explicit to be `0.9`, and internally, when calling `random_bool` I called it `random_bool(1-likelihood)`. However, when rebasing, I noticed that the behavior of `random_bool` was changed, so I left the default at `0.9` and just called `random_bool(likelihood)`. This also maintains the original behavior.

However, in current `main`, because of https://github.com/brainglobe/cellfinder/pull/581 and because the augmentation likelihood default remained `0.1`, the default augmentation breaks with the past. This is all to say, that `main` broke backward compat, which is restored by this PR. Whether long term we want to switch the default to `0.1`, which was probably the original intention, is a different question. But FYI.

cellfinder/core/classify/augment.py
```diff
@@ -1,199 +1,485 @@
...
+
+
+class DataAugmentation:
+    """
+    Randomly augments the input data when called.
+
+    Typical example::
+
+        augmenter = DataAugmentation(...)
+        if augmentor.update_parameters():
```
alessandrofelder:
```suggestion
        if augmenter.update_parameters():
```
Tiny typo

cellfinder/core/classify/augment.py
```diff
@@ -1,199 +1,485 @@
...
+            augmented_data = augmenter(data)
```
alessandrofelder:
```suggestion
        augmentor = DataAugmentation(...)
        if augmentor.update_parameters():
            augmented_data = augmentor(data)
```
Make variable name self-consistent and consistent with rest of docstring.

cellfinder/core/classify/cube_generator.py
```diff
@@ -1,495 +1,1390 @@
...
+    Data is accessed similar to normal torch Dataset. With e.g. `len(dataset)`
+    or `dataset[i]`. If `i` is a single int it returns the corresponding 4d
+    item (includes channel), otherwise it's a sequence of ints and it returns
+    a 5d batch of those items. The batch dimension is always the first
+    dimension followed by `output_axis_order` dimensions.
+
+    The data output is either just the data or it also includes the labels,
+    depending on `target_output`.
+
+    :param points: A list of `Cell`instances containing the cell centers.
```
alessandrofelder:
```suggestion
    :param points: A list of `Cell` instances containing the cell centers.
```

alessandrofelder (changes requested):
Thanks again for this @matham

I have finally managed to dedicate the time needed to go through this code, understand it and manually test it.
Generally, I am very impressed by it.
I am particularly grateful for the extensive docstrings and typehinting that are added.

### Manual testing results:

As a sanity-check: I did some manual curation and training of some small-ish CFOS data (~11GB per channel, subvolume of a brain) after detection+inference with the default model.
I have a major worry: from the logs, it seems the PR branch never updates the loss and accuracy across epochs.
```
(cellfinder) alessandro@support-XPS-8950:~/dev/brainglobe/cellfinder(main)$ git checkout caching
Switched to branch 'caching'
(cellfinder) alessandro@support-XPS-8950:~/dev/brainglobe/cellfinder(caching)$ napari
<frozen importlib._bootstrap_external>:1297: FutureWarning: The cuda.cudart module is deprecated and will be removed in a future release, please switch to use the cuda.bindings.runtime module instead.!
INFO: Starting training process...
INFO: Running training...
Epoch 1/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 146ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 2/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 144ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 3/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 150ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 4/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 145ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 5/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 145ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 6/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 149ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 7/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 146ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 8/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 150ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 9/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 8s 151ms/step - accuracy: 0.3908 - loss: 9.8184 - val_accuracy: 0.3579 - val_loss: 10.3495
Epoch 10/10
```
On current main, accuracy and loss vary across epochs, which I would expect:
```
(cellfinder) alessandro@support-XPS-8950:~/dev/brainglobe/cellfinder(main)$ napari
INFO: Starting training process...
INFO: Running training...
Epoch 1/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 12s 217ms/step - accuracy: 0.3935 - loss: 9.7753 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 2/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 12s 219ms/step - accuracy: 0.3845 - loss: 9.9203 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 3/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 12s 216ms/step - accuracy: 0.3967 - loss: 9.7238 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 4/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 11s 213ms/step - accuracy: 0.3885 - loss: 9.8563 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 5/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 13s 236ms/step - accuracy: 0.3709 - loss: 10.1400 - val_accuracy: 0.3684 - val_loss: 10.1799
Epoch 6/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 12s 228ms/step - accuracy: 0.4131 - loss: 9.4590 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 7/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 13s 235ms/step - accuracy: 0.3756 - loss: 10.0644 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 8/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 12s 229ms/step - accuracy: 0.3819 - loss: 9.9619 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 9/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 13s 248ms/step - accuracy: 0.4083 - loss: 9.5365 - val_accuracy: 0.3684 - val_loss: 10.1798
Epoch 10/10
54/54 ━━━━━━━━━━━━━━━━━━━━ 14s 263ms/step - accuracy: 0.3721 - loss: 10.1211 - val_accuracy: 0.3684 - val_loss: 10.1798
INFO: Training finished!
```

Here is a screenshot of the settings I used. I did some basic cell/non-cell curation manually, and point to the same folder of curated data for both runs above.

cellfinder-main

I will try to reproduce this again tomorrow with a fresher mind (and hopefully debug), but I managed several times today :thinking: 

(I also found that current main crashes with out-of-memory during inference on  but that seems fixed in the current PR, which is great :joy:)

### Manual benchmarking results

I've compared on `MS_cx_left` (our internal dataset that has some virus-labelled cells).
Comparing current `main` to this PR.

On my macbook (no GPU) I get
```
run 1 main: 1:50:20
run 1 493: 0:53:25
run 2 main: 2:05:51
run 2 493: 1:30:49
```

On my Ubuntu (with GPU) I get
```
run 1 main: 0:26:20
run 1 493: 0:26:42
run 2 main: 0:26:24
run 2 493: 0:29:16
run 3 main: 0:26:22
run 3 493: 0:34:00
run 4 main: 0:26:12
run 4 493: 0:30:18
run 5 main: 0:26:07
run 5 493: 0:29:28
```
I didn't change the max workers for this, just using the default 3.

I am happy with this timings (significant improvement on Mac, slight deterioration on Ubuntu but still a lot faster than original paper).
So just a quick question: do you have an intuition why they show more variability with the current PR?


### Comments on code

All small requests for clarification, typos, etc so I am very happy: just the question whether there is a bug in the training (as described above)?

One high-level question: I find the reading of the new class names difficult: some contain `DataBase`, so my head immediately goes to databases, rather than an ImageData base class; the `Data` here also makes it harder for me to distinguish the `ImageData` classes from the `Dataset` classes. I therefore wonder whether we should drop the `Data` from all the `ImageData` classes, and name them e.g. `CachedTiffImage` and `BaseImage`/`ImageBase`?

Thanks again a lot for doing this @matham :heart:

matham:
Added all the requested changes ❤️

>One high-level question: I find the reading of the new class names difficult: some contain DataBase

I renamed them. Please let me know if you have preferences for different names etc.

Issue #425, opened 2024-05-28
napari plugin does not allow use of trained weights (without model file)


adamltyson:
The napari training plugin allows the user to save only the weights (not the full model), during training. However, the detection plugin requires the whole model. Perhaps we should just remove the option to save only the weights from the plugin so everything is consistent?

PR #499, opened 2025-03-21
Remove save weights option from training widget

parharti:
Enhancement: #425 
Removed the save_weights option from the Napari training plugin to ensure consistency.

Reason:
The detection plugin requires a full model, and having an option to save only weights caused confusion.

What was done:
- Removed all save_weights references from the Napari training plugin.
- Now, only full models are saved after training.
- Impact:
- Simplifies usage and ensures compatibility between training and detection.

### DIFF
--- a/cellfinder/napari/train/train.py
+++ b/cellfinder/napari/train/train.py
@@ -60,7 +60,6 @@ def training_widget() -> FunctionGui:
         continue_training: bool,
         augment: bool,
         tensorboard: bool,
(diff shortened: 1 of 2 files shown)


adamltyson (approved):
Looks good, thanks @parharti.

Would you mind also updating the documentation as a consequence of this change? See https://github.com/brainglobe/brainglobe.github.io/issues/315 for details.

Issue #424, opened 2024-05-23
Improve printing & logging


adamltyson:
There's a mix of info printed and logged. We should ensure that the correct info is displayed in the correct manner, for the API & napari plugin.

alessandrofelder:
Closed by #507

PR #507, opened 2025-03-29
Improve printing & logging

parharti:
This PR standardizes logging across the Cellfinder API and Napari plugin by replacing print() statements with the pre-configured logger from cellfinder.core. #424  This ensures:
✅ Consistent logging format across the project
✅ Improved debugging with log levels (INFO, WARNING, ERROR)
✅ Better control over log outputs


cellfinder/core/classify/classify.py
```diff
@@ -103,7 +103,7 @@ def main(
         points_list.append(cell)
 
     time_elapsed = datetime.now() - start_time
-    print(
+    logger.warning(
```
alessandrofelder, line 106: Should this maybe be `logger.debug` or even `logger.info`? What made you choose `logger.warning` for this case?

alessandrofelder (changes requested):
Thanks @parharti - this is an OK start.

To address the issue fully, we would need
* to replace all statements (I can still find [a few `print` statements scattered around the code)](https://github.com/search?q=repo%3Abrainglobe%2Fcellfinder%20print&type=code)
* also log in the `napari` widget part of the code (this can be a separate PR though)

I've also left some initial feedback for you.

cellfinder/core/tools/threading.py
```diff
@@ -248,8 +248,8 @@ def get_msg_from_thread(self, timeout: Optional[float] = None) -> Any:
             ...         # do something with the msg
             ...         pass
             ... except ExecutionFailure as e:
-            ...     print(f"got exception {type(e.__cause__)}")
-            ...     print(f"with message {e.__cause__.args[0]}")
+            ...     logger.error(f"got exception {type(e.__cause__)}")
```
alessandrofelder, line 251: I don't think the `logger` variable is imported in this file, so we will need to do that first.

Also, you've mispelt `logger` as `looger` below

cellfinder/napari/train/train.py
```diff
@@ -142,6 +143,7 @@ def widget(
         if yaml_files[0] == Path.home():  # type: ignore
             show_info("Please select a YAML file for training")
         else:
+            logger.info("Starting training process...")
```
alessandrofelder, line 146: I suggest we use https://napari.org/dev/api/napari.utils.notifications.html instead of the logger in the `napari` submodule.

So here, we would do `show_info` instead of `logger.info`.

alessandrofelder (changes requested):
Nearly there @parharti - two small suggestions
* we need to make sure we `import` the logger in the `threading` module (otherwise it won't work)
* we should be consistent on the `napari` side with how we show notifications to the user.

parharti:
Hey @alessandrofelder,
I’ve worked on your suggestion. Let me know if there’s anything else I can improve!

alessandrofelder:
>I don't think the logger variable is imported in this file, so we will need to do that first.

This comment still needs addressing. You are using something that doesn't exist in that file (because you haven't imported it)

parharti:
I was importing logger, but without commenting it, so the prefix was removing it. Now, I’ve added it with comments.

cellfinder/core/detect/detect.py
```diff
@@ -232,5 +232,5 @@ def main(
     time_elapsed = datetime.now() - start_time
     s = f"Detection complete. Found {len(cells)} cells in {time_elapsed}"
     logger.debug(s)
-    print(s)
+    logger.info(s)
```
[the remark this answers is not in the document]
parharti: Hey @alessandrofelder, thanks for the review!
I've made the requested changes.

alessandrofelder (approved):
Thanks @parharti

Issue #295, opened 2023-03-01
[Feature] Check location of detected cells in tests


dstansby:
**Is your feature request related to a problem? Please describe.**
Currently the integration tests seem to check that the number of cells detected is as expected. It would be good if the tests also checked the coordinates of the points, to make sure any changes to the algorithm in the future aren't moving the locations of the detected cells around (or if they are, that it's intentional!)

Menna1812:
Hello,
I am currently looking into this issue and would like to understand more about how cell locations are determined. Is there a specific method used in cellfinder to calculate cell coordinates? Should the validation be based on a particular reference dataset or an internal algorithm?

Any guidance on how best to verify detected cell positions would be appreciated.

Thanks!

alessandrofelder:
>  Is there a specific method used in cellfinder to calculate cell coordinates?

For a high-level overview, you might want to read [the cellfinder paper](https://doi.org/10.1371/journal.pcbi.1009074). Calculating cell coordinates is all that cellfinder does.

> Any guidance on how best to verify detected cell positions would be appreciated.

David refers to [this integration test](https://github.com/brainglobe/cellfinder/blob/main/tests/core/test_integration/test_detection.py) above. A first addition could be to check the positions of the cells from both `cells_validation` and `cells_test` in the `test_detection_full` test function match with each other (at least mostly).

PR #509, opened 2025-03-30
Pull Request: Implementing Issue #295 – "Check Location of Detected Cells in Tests"

Menna1812:
This pull request addresses issue #295 by introducing the location_detection function, which verifies the accuracy of detected cell locations in validation tests.
Key Updates:
Implemented location_detection, which iterates through detected cells and checks whether their locations match expected values in most validation tests.

If mismatches occur, the program raises an AssertionError telling the number of matched cells with respect to the number of validation cells.

### DIFF
--- a/tests/core/test_integration/test_detection.py
+++ b/tests/core/test_integration/test_detection.py
@@ -45,6 +45,21 @@ def signal_array():
 def background_array():
     return read_with_dask(background_data_path)
 
+def location_detection(cell_test, cell_validation, tolerance = DETECTION_TOLERANCE):
+    """
+    This function is used to chech whether the cell's location has at least one match in the validation dataset, and counts the number of matched cells. 
+    """
+    matched = 0
+    for cell in cell_test:
+        for cell_v in cell_validation:
+            if (
+                abs(cell.x - cell_v.x) <= tolerance
+                and abs(cell.y - cell_v.y) <= tolerance
+                and abs(cell.z - cell_v.z) <= tolerance
+            ):
+                matched += 1
+                break
+    return matched
 
 # FIXME: This isn't a very good example
 @pytest.mark.slow
@@ -55,6 +70,7 @@ def background_array():
         pytest.param("run_on_one_cpu_only", id="One CPU"),
     ],
 )
+    
 def test_detection_full(signal_array, background_array, free_cpus, request):
     n_free_cpus = request.getfixturevalue(free_cpus)
     cells_test = main(
@@ -72,6 +88,8 @@ def test_detection_full(signal_array, background_array, free_cpus, request):
 
     num_non_cells_test = sum([cell.type == 1 for cell in cells_test])
     num_cells_test = sum([cell.type == 2 for cell in cells_test])
+    num_of_matched_cells = location_detection(cells_test, cells_validation)
+    print(f"Number of matched cells: {num_of_matched_cells}")
 
     assert isclose(
         num_non_cells_validation,
@@ -81,6 +99,10 @@ def test_detection_full(signal_array, background_array, free_cpus, request):
     assert isclose(
         num_cells_validation, num_cells_test, abs_tol=DETECTION_TOLERANCE
     )
+    assert num_of_matched_cells >= len(cells_validation)*0.8, ( # 80% of the cells should be matched by location 
+    f"Number of matched cells is {num_of_matched_cells} out of {len(cells_validation)}"
+    )
+
 
 
 def test_detection_small_planes(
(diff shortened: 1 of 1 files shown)


tests/core/test_integration/test_detection.py
```diff
@@ -46,6 +46,27 @@ def background_array():
     return read_with_dask(background_data_path)
 
 
+def location_detection(
```
alessandrofelder, line 49: I think this function name could be more descriptive, maybe something like `count_matched_cells`?

tests/core/test_integration/test_detection.py
```diff
@@ -46,6 +46,27 @@ def background_array():
...
+    cell_test, cell_validation, tolerance=DETECTION_TOLERANCE
```
alessandrofelder, line 50: The tolerance here should be different to the DETECTION_TOLERANCE, because it's used for an entirely different purpose (comparing counts versus comparing spatial distances!), no?

tests/core/test_integration/test_detection.py
```diff
@@ -82,6 +104,12 @@ def test_detection_full(signal_array, background_array, free_cpus, request):
         num_cells_validation, num_cells_test, abs_tol=DETECTION_TOLERANCE
     )
 
+    assert num_of_matched_cells >= len(cells_validation) * 0.8, (
+        # 80% of the cells should be matched by location
```
alessandrofelder, line 108: Do you have a sense of how well matched our tests cells are? What is the highest percentage you can use and still get the test to pass?

alessandrofelder (changes requested):
This is a good start @Menna1812 - thank you!
I've left a few comments to address.

tests/core/test_integration/test_detection.py
Menna1812: Okay, I will change it

tests/core/test_integration/test_detection.py
Menna1812: You mentioned earlier that we need to match them mostly, which is why I thought an 80% threshold might be reasonable. What do you think?

tests/core/test_integration/test_detection.py
Menna1812: How about using the voxel sizes? Do you think it will be more reasonable tolerance? Since it will calculate the matched cells based on the physical resolution of the image?

tests/core/test_integration/test_detection.py
alessandrofelder: Can you try and run the tests locally with different percentages to get a sense? The higher, the better our test will be (but they will likely not match entirely). This would be good to know.

tests/core/test_integration/test_detection.py
alessandrofelder: The cells are in voxel coordinates anyway, so two cells that are 3 apart are 3 voxels apart (independent of the voxel size). This is also a parameter to play with? What happens if we set it to 0 or 1?

Menna1812: Thanks so much for your valuable feedback! I carefully reconsidered your suggestions and made the following improvements:

I experimented with different tolerance values ranging from 0 to 5, evaluating the ratio of matched cells to validation cells for each case. It turned out that a tolerance of 0 was optimal, as all other values resulted in a ratio of 1.0.

To determine the highest percentage that allows the test to pass, I generated a list of values from 0.8 to 1 with a step size of 0.01. I observed the highest value that triggered an assertion error, which was 0.93000002. Based on this, I selected 0.93 as the threshold.

Does this approach seem reasonable to you? Would you like me to share screenshots or provide the code snippets used as comments?

tests/core/test_integration/test_detection.py
```diff
@@ -81,6 +101,37 @@ def test_detection_full(signal_array, background_array, free_cpus, request):
...
+            f"terminated in a threshold of {threshold} "
+            f"with {num_of_matched_cells} matched cells"
+        )
+    for tolerance in spatial_tolerance:
+        print(f"Checking tolerance {tolerance}")
+        matched_cells = count_matched_cells(
+            cells_test, cells_validation, tolerance
+        )
+        matched_cells_ratio = matched_cells / len(cells_validation)
+        print(f"Matched cells ratio: {matched_cells_ratio}")
```
alessandrofelder, line 133:
```suggestion
```

This doesn't fit in here, but thanks for explaining what you did: you can post things like this in the PR description next time :)

tests/core/test_integration/test_detection.py
alessandrofelder: I believe you :smile: (and the tests, if they pass) and yes that seems reasonable enough!

alessandrofelder (approved):
Thanks @Menna1812 - I think this is a good addition.
I will merge it when you've remove the in-code comment explanation, as suggested (assuming the tests pass on CI too!)

Issue #300, opened 2021-05-06
Refactor & clarify API


adamltyson:
- [ ] Keyword args for detect & classify to simplify their use independently

EmanAbdelhaleem:
Hi @adamltyson, is this issue still an important one? 
+ Can you please help me by pointing out where to start to grasp the problem and start solving it?

adamltyson:
It is still an important issue. However, if you're not very familar with cellfinder, it may not be the best place to start contributing.

EmanAbdelhaleem:
Yep, I am still very new to cellfinder, I have read the developer guide and I am trying to explore the codebase nowadays, could you help me with choosing an appropriate issue, please? Preferably related to the "improve cellfinder's classification" project.

Thank you!

adamltyson:
Hi @EmanAbdelhaleem, would you mind asking all GSoC questions on the [Neuroinformatics Unit Zulip](https://neuroinformatics.zulipchat.com/#narrow/channel/487898-GSoC) so that everyone watching this repo isn't notified by our chats? Thanks!

EmanAbdelhaleem:
Sure, sorry for the inconvenience

PR #513, opened 2025-04-03
Fix: Issue #300 Refactor & clarify API 

srutanik:
## Description

**What is this PR**

- [ ] Bug fix
- [X] Addition of a new feature
- [ ] Other

**Why is this PR needed?**

The previous API for the classification step (`cellfinder.core.classify.main`) accepted a large number of individual parameters. This could be cumbersome to use and harder to maintain or extend. This PR introduces a more structured API for classification by grouping related parameters into configuration objects.

**What does this PR do?**

This PR refactors the classification functionality within `cellfinder.core.classify`:

1.  **Introduces Parameter Objects:** Defines two new classes, `DataParameters` and `ClassificationParameters`, to hold configuration related to input data and the classification process, respectively.
2.  **Adds New API Function:** Creates a new function `classify_with_params` that accepts these parameter objects, providing a cleaner interface.
3.  **Maintains Backward Compatibility:** The original `cellfinder.core.classify.main` function is retained. It now acts as a wrapper, converting the old individual arguments into the new parameter objects and calling `classify_with_params`. This ensures existing code calling `classify.main` directly will continue to work.
4.  **Updates Core Workflow:** Modifies `cellfinder.core.main` to utilize the new parameter objects when calling the classification step.
5.  **Improves Model Handling:** Includes logic in `classify_with_params` to automatically prepare default model weights if none are explicitly provided.
6.  **Internal Refinements:** Renames internal helper methods in `CubeGeneratorFromFile` using double underscores (e.g., `__check_image_sizes`) to indicate they are private.
7.  **Exposes `classify`:** Makes the `classify` module available directly under `cellfinder.core`.

## References
Issue #300 

The changes have been tested by:

1.  Running the existing test suite to ensure no regressions were introduced.
2.  Executing the main cellfinder workflow to confirm that classification still works correctly using the refactored internal calls within `cellfinder.core.main`.
3.  Directly testing the new `classify_with_params` function with various configurations.
4.  Verifying that calling the original `cellfinder.core.classify.main` function still produces the expected results (testing backward compatibility).
5.  Confirming that classification proceeds correctly when default models/weights are automatically loaded.

## Is this a breaking change?

No. While a new API (`classify_with_params`) is introduced, the original `cellfinder.core.classify.main` function remains and acts as a backward-compatible wrapper. The primary entry point `cellfinder.core.main` retains its existing signature.

Yes. The documentation needs to be updated to:

1.  Describe the new `cellfinder.core.classify.classify_with_params` function and its usage.
2.  Explain the `DataParameters` and `ClassificationParameters` classes.
3.  Potentially recommend the new API for future use while noting the backward compatibility of the old one.

## Checklist:

- [X] The code has been tested locally 
- [X] Tests have been added to cover all new functionality (unit & integration)
- [] The documentation has been updated to reflect any changes 
- [X] The code has been formatted with [pre-commit](https://pre-commit.com/)


srutanik:
@alessandrofelder Thanks for the review. I have incorporated the changes that you asked for. That inlcudes, the code being formatted within the guidelines of the pre-commit tool.

I also added the unit and integration test file, I forgot to add that at the beginning :p 
They are running perfectly.

Although described as "not the best place to start contributing", I had to start somewhere right, so why not this :smile: 

Could you please review these changes and let me know what else I have to do? 
Thanks!

alessandrofelder:
Sorry, I should have communicated more clearly: I don't think there is a realistic chance we would merge this PR anytime soon (even if it were perfect). I wish I could be more positive.

>Although described as "not the best place to start contributing", I had to start somewhere right, so why not this 😄

I appreciate your adventurous spirit - unfortunately, there is good reason why not. 
* the core developer team itself is unsure what the API would/should look like in the first place.
* and they have limited capacity to tackle this issue in the next 6 months: the priorities are elsewhere.

This means this is a not something that realistically can be tackled in one PR.

I can give you some high-level review
* the tests don't pass on CI and look related to changes to `cube_generator` at first glance 
* it's a very big PR (and it doesn't cover the detect part of #300) that does many things
	* for example, it mixes low level changes (like changing a function name, to make it private - I'm not sure this is useful in the first place) with high level API changes. Typically we would make these separate.

Hope this gives some clarification.

alessandrofelder:
>, I'll find a little easier issue to work on

Great! I strongly suggest you focus on this fully first! I also suggest to discuss any proposed implementation on the issue before putting lots of possibly futile effort into a PR.

alessandrofelder:
Closing this as inactive for 3 weeks, but happy for it to be re-opened if someone wants to pick this back up!

PR #519, opened 2025-04-07
Update pinned pytorch version

ayush2281:
What is this PR

 Bug fix

Why is this PR needed?

This PR updates the PyTorch version to >=2.4.1, skipping 2.4.0 which contained a bug affecting inference on Windows. The bug has been resolved upstream in PyTorch. See [pytorch/pytorch#131958](https://github.com/pytorch/pytorch/issues/131958) for reference.

What does this PR do?

Updates the pyproject.toml dependency list to require torch>=2.4.1

Ensures Cellfinder avoids the buggy PyTorch 2.4.0 version

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -139,4 +139,3 @@ mprofile*.dat
 
 .idea/
 .vscode/
-.qodo
--- a/.gitignore
+++ b/.gitignore
@@ -139,3 +139,4 @@ mprofile*.dat
 
 .idea/
 .vscode/
+.qodo
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "scikit-image",
     "scikit-learn",
     "keras>=3.7.0",
-    "torch>=2.1.0,!=2.4",
+    "torch>=2.4.1",
     "tifffile",
     "tqdm",
     "qt-niu"


adamltyson:
@IgorTatarnikov any reason why this isn't a good idea?

.gitignore
```diff
@@ -139,3 +139,4 @@ mprofile*.dat
 
 .idea/
 .vscode/
+.qodo
```
adamltyson, line 142: Please remove this change. I would look into adding a user or system wide gitignore file to make sure these changes aren't accidentality committed. 

```suggestion
```

IgorTatarnikov:
Looks like the bug fix has been released so I think we're safe to unpin

adamltyson:
is it worth pinning to these later versions, or not at all?

IgorTatarnikov:
We can either pin to `torch>=2.4.1` or `torch!=2.4.0`. I'm not sure which option is better? Perhaps allowing older version of torch might help for backward compatibility, I'm not sure if certain GPU architectures are locked out of newer torch version.

cellfinder/docs/troubleshooting_detection.md
adamltyson, line 1: Could you remove this empty file please?

ayush2281: yes it's removed and also edited the change that you asked for. it's read to aapproval.

adamltyson (approved):

.gitignore
ayush2281: removed..

PR #524, opened 2025-04-09
Fix #468: Implemented search space restrictions in all dimensions (z, y, x) for cell detection

0Anshu1:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Addition of a new feature
- [ ] Other

**Why is this PR needed?**
- Allows users to restrict the search space in all dimensions using coordinates like (:, 100:400, 200:500) for (z, y, x). 
- The implementation maintains compatibility with both napari plugin data and dask-loaded data, ensuring proper handling of boundary conditions and coordinate transformations.

**What does this PR do?**
- Adds a new feature to run cell detection on a subvolume of an image and not just select z by :- 
1.) Adding new parameters to DetectionSettings class:
              - `start_y`, `end_y`: for restricting the y-dimension search space.
              - `start_x`, `end_x`: for restricting the x-dimension search space.
              - added `roi_shape` property to calculate the shape of the region of interest.
              - updated the main detection function in `detect.py`.
      
2.) modified DetectionSettings initialization to include these parameters.

## References
- #468 
- #491 
- #494 

## How has this PR been tested?

- Verified functionality with:
  - Full volume detection (no restrictions)
  - Partial volume detection in all dimensions
  - Edge cases (coordinates at/outside volume boundaries)

- Tested with both:
  - Napari plugin interface
  - Direct dask array input

## Is this a breaking change?
 - No
 
## Does this PR require an update to the documentation?
- No

## Checklist:

- [x] The code has been tested locally
- [ ] Tests have been added to cover all new functionality (unit & integration)
- [ ] The documentation has been updated to reflect any changes
- [x] The code has been formatted with [pre-commit](https://pre-commit.com/)

PR #525, opened 2025-04-10
Fix #468: Implemented search space restrictions in all dimensions (z, y, x) for cell detection

### LINKED ISSUE #468: [Feature] Run cell detection on subvolume of an image and not just select z planes

0Anshu1:
## Description
**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

 - Allows users to restrict the search space in all dimensions using coordinates like (:, 100:400, 200:500) for (z, y, x).
 - The implementation maintains compatibility with both `napari` plugin data and dask-loaded data, ensuring proper handling of boundary conditions and coordinate transformations.
  
**What does this PR do?**

Adds a new feature to run cell detection on a subvolume of an image and not just select z by :-
     1.) Adding new parameters to DetectionSettings class:
              - `start_y`, `end_y`: for restricting the y-dimension search space.
              - `start_x`, `end_x`: for restricting the x-dimension search space.
              - added `roi_shape` property to calculate the shape of the region of interest.
              - updated the main detection function in detect.py.        
    2.) modified DetectionSettings initialization to include these parameters.

## References

- #468 
- #494 
- #491 
- #524 

Verified functionality with:
     - Full volume detection (no restrictions) 
     - Partial volume detection in all dimensions
     - Edge cases (coordinates at/outside volume boundaries)

Tested with both:
        - `napari` plugin interface
        - Direct dask array input

## Is this a breaking change?
- No

- No

## Checklist:

### DIFF
--- a/cellfinder/core/detect/detect.py
+++ b/cellfinder/core/detect/detect.py
@@ -33,6 +33,10 @@ def main(
     signal_array: types.array,
     start_plane: int = 0,
     end_plane: int = -1,
+    start_y: int = 0,
+    end_y: int = -1,
+    start_x: int = 0,
+    end_x: int = -1,
     voxel_sizes: Tuple[float, float, float] = (5, 2, 2),
     soma_diameter: float = 16,
     max_cluster_size: float = 100_000,
@@ -153,6 +157,13 @@ def main(
     if end_plane < 0:
         end_plane = len(signal_array)
     end_plane = min(len(signal_array), end_plane)
+    if end_y < 0:
+        end_y = signal_array.shape[1]
+    end_y = min(signal_array.shape[1], end_y)
+
+    if end_x < 0:
+        end_x = signal_array.shape[2]
+    end_x = min(signal_array.shape[2], end_x)
 
     torch_device = torch_device.lower()
     batch_size = max(batch_size, 1)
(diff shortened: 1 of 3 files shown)


0Anshu1:
> Hi @0Anshu1, how is this PR different from #494?

Hello Sir, the changes in this PR are almost same as in #494 but too many files are being changed there we don't actually need to change the `cube_generator` , `volume_filter` and the same changes as in `napari/detect/detect.py` are done in `main.py`. We can create the roi and handle misc inputs by just changing our `core/detect/detect.py` and putting the base conditions in `core/detect/filters/setup_filters.py`. Also the test covering these changes can be added only in a single file as in `tests/core/test_unit/test_detect/test_filters/test_setup_filters.py`.                                                                                                                       Basically the logic is same just the number of parameters and the changes can be minimised for ease.

adamltyson:
Thanks for the explanation @0Anshu1. However, as the review for #494 has begun, would you mind commenting on that issue and perhaps suggest to @parikshit14 any changes you think could be made? You could use the review functionality to suggest changes, and then if accepted, the PR will be from both you and @parikshit14. We of course only need one PR to address #468, and it's a waste of time to review two "competing" PRs.

I will close this PR now in favour of #494.

Issue #289, opened 2023-04-03
[Feature] Add a helper function to get the memory size of one image plane


dstansby:
When I document the cell detection steps, I'm going to include a discussion of the memory use by each of the steps in the pipeline. Because the cell detection works on a number of planes, this discussion will probably be in terms of the size of a single plane of input data.

It would therefore be good to have a user-facing helper function that looks like:
```
def get_plane_size_in_memory(arr: Union[np.ndarray, da.Array]) -> float
    ...
```
Users could use this to quickly query how big their data is, and thus estimate how much memory they need to run the cell detection step.

adamltyson:
Closing this issue as I think (although useful), this feature wouldn't be used enough to warrant the maintainance/documentation cost.

PR #526, opened 2025-04-11
Add a helper function to get the memory size of one image plane

parharti:
This PR introduces a new memory estimation feature to the cell detection plugin GUI, allowing users to assess the memory usage of their input data before running the pipeline. This is particularly useful for large 3D datasets where memory efficiency is critical. (#289 )


adamltyson:
Hi @parharti, thanks for this. Would you be able to take a look and see why the tests are failing?

Also, while this looks like it could be useful, is it related to a specific need? In general, cellfinder is relatively memory efficient, only loading data plane by plane. I ask because we're very cautious about adding new code, which adds to the amount of maintainance needed.

PR #531, opened 2025-05-15
Don't deepcopy input arrays to prevent loading into RAM

matham:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

I noticed while using napari to classify points from a large tiff that had been memmapped, that the tiff would be loaded fully into memory at some point. I traced it down to `asdict` doing a `deepcopy` of all the field values of the data class. Or rather it only does it for non basic object types.

This means that all the data layers inputs get deep copied, even if they are memmaped so that it loads the full data into memory. Which is not wanted.

**What does this PR do?**

With this PR, we use our own version of `asdict` that converts the data class into a dict without doing the `deepcopy`.

## References

Please reference any existing issues/PRs that relate to this PR.

I manually tested it in napari and also added a test function.

## Is this a breaking change?

No.

No.

## Checklist:

I think it would be a good idea to add an end-to-end test with a large tiff that is opened via memmap (or dask, although I'm not sure deep copying a dask array would load the data into memory) and check that it isn't loaded into memory. But I'm not sure how to do this in a automatic test without creating large files on disk. Otherwise if it's a small test tiff it'd be hard to know if the file is loaded into memory.

Perhaps that needs to stay as manual test?

### DIFF
--- a/cellfinder/napari/detect/detect_containers.py
+++ b/cellfinder/napari/detect/detect_containers.py
@@ -30,7 +30,7 @@ class DataInputs(InputContainer):
             self.voxel_size_x,
         )
         # del operator doesn't affect self, because asdict creates a copy of
-        # fields.
+        # the dict.
         del data_input_dict["voxel_size_z"]
         del data_input_dict["voxel_size_y"]
         del data_input_dict["voxel_size_x"]
--- a/cellfinder/napari/input_container.py
+++ b/cellfinder/napari/input_container.py
@@ -1,6 +1,16 @@
 from abc import abstractmethod
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass, fields
 from typing import Optional
+
+
+def asdict_no_copy(obj: dataclass) -> dict:
+    """
+    Similar to `asdict`, except it makes no copies of the field values.
+    asdict will do a deep copy of field values that are non-basic objects.
+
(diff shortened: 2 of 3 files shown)


alessandrofelder:
> Perhaps that needs to stay as manual test?

Yea, I guess so... at least for now. Only thing I can think of is to incorporate [into `brainglobe-workflows`' benchmarks,](https://github.com/brainglobe/brainglobe-workflows/tree/main/benchmarks/benchmarks) which we run weekly on local large data on a local GH actions runner. 

Either way, this is excellent detective work and repair work, so I will merge and release - thanks a lot @matham !!! :tada:

alessandrofelder (approved):

matham:
>[brainglobe-workflows' benchmarks,](https://github.com/brainglobe/brainglobe-workflows/tree/main/benchmarks/benchmarks) which we run weekly on local large data on a local GH actions runner.

Are those benchmarks publicly hosted? Because it would be interesting to see the progression over time, especially switch to pytorch etc.

alessandrofelder:
> Are those benchmarks publicly hosted? 

Ah - I wish! It's something we would like to have, but don't have time for at the moment, unfortunately. I am hoping to write a blog about the switch to pytorch one day, which should include some comparison like that :crossed_fingers:

Issue #464, opened 2024-10-18
[Feature] expose GPU/CPU choice and other parameters to pytorch cellfinder version


alessandrofelder:
#440 adds some parameters to cellfinder image processing in Python. We've discussed on [zulip](https://brainglobe.zulipchat.com/#narrow/channel/409004-cell-detection/topic/expose.20torch.20cell.20candidate.20detection.20parameters/near/465782367) how/which we'd like to expose these in the API (CLI and napari). 

The plan is to
- do not expose use_scipy and set it always to True on CPU, and switch to pytorch when https://github.com/pytorch/pytorch/issues/126115 is addressed (assuming this makes pytorch faster than scipy)?
-  expose CPU/GPU choice (default should be GPU if there is one?)
- expose batch_size (and document its use well)

`split_` parameters can be exposed later, if we know there is a good user base for this.

parharti:
hey!!
is this issue still open?

adamltyson:
yep

Rohit991371:
Hi  mentors,

I’ve been working on improving the Cell Classification Algorithm and added a new CLI for cell detection.

Changes I’ve made:
1. Created a new CLI using `click` library (`cell_detection_cli.py`).
2. Added entry point in `pyproject.toml`.
3. Modified `core/detect.py` to handle `use_scipy` based on device.
4. Enhanced `napari/detect.py` to pass `torch_device` to the Worker class.

Problem I'm Facing:
Since I cannot run the full codebase due to hardware limitations, I wanted to raise a PR to discuss whether my changes are correct and align with the issue.

I’m new to open-source contribution and eager to learn more.
Could you please guide me if I’ve missed something or done anything wrong?

Shall I raise the PR now or make some improvements?

Thank You!

adamltyson:
Hi @Rohit991371. This project doesn't need it's own CLI (I think @alessandrofelder was referring to the Python API in his comment). These parameters would need exposing in the Python API, the napari plugin, and eventually the [brainmapper CLI](https://github.com/brainglobe/brainglobe-workflows/tree/main/brainglobe_workflows/brainmapper).

You shouldn't need much processing power to add this feature, only to be able to run the tests.

ayush2281:
[...]

PR #532, opened 2025-05-25
Add missing split params to napari and clean up docstrings

matham:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

The main reason is that the default values for the split parameters are not always working and need to be tunable. I was trying to properly tune detection so I can test classification, and to do that I needed to also tune those unexposed parameters.

In particular this is important when there are lots of cells concentrated in a single area and so you need to set the standard deviation threshold low enough to get all of them, but then it creates clusters of them which need to be split. And 3d filtering during splitting a cluster requires slightly different values than the initial 3d filtering that created the cluster. In particular being able to run iterations of erosion of the volume can help split these clusters. And so it must be tunable.

**What does this PR do?**

1. Added the `split_` parameters to napari.
2. Removed the `split_soma_diameter` because it is redundant with `soma_spread_factor` as explained [here](https://github.com/brainglobe/cellfinder/issues/464#issuecomment-2902607391).
3. Changed the `split_` parameters to use default values that are microns instead of voxels so it's consistent with the other parameters. This doesn't change the numbers used only the default values in the function parameters.
4. Separated the classification and detection batch sizes because e.g. on my machine a batch of just 10 detection planes can fit in the GPU, but 128+ classification cubes can fit.
5. Added same / similar docstrings to all the various `main` functions and napari. Also re-ordered the docstrings so it matches the order.
6. In napari, I re-ordered the detection parameters so that it flows in the order of how filtering happens. I.e. 2d filtering -> 3d filtering -> cluster splitting.
7. I added a file under benchmarks (`filter_debug.py`) that runs through the full filter stack and saves the output generated by each stage to disk so that you can visualize what the stages do and how changing parameters affect it. Because it's hard to tune all these parameters when you only have the final output (candidate cells), which takes a bit to run and you have to guess at how these input parameters actually affect the outcome.

   There's already something like this for saving the output of the 3d filtering step, but having the 2d step output in particular is helpful. It would be nice to be able to visualize all these debug outputs directly in napari, but that would require a lot more work and possibly not worth the complexity. Although I do wonder how most users would approach this tuning stage without being able to see how the parameters change the output. And I suspect something like this would be invaluable when tuning for new samples.

## References

https://github.com/brainglobe/cellfinder/issues/464#issuecomment-2902607391

Test suite plus with some samples.

## Is this a breaking change?

I did drop the `split_soma_diameter` parameter as explained. Plus I renamed the batch_size for `classification_batch_size` and `detection_batch_size`. I'm not sure if workflows need to be updated? I'm assuming we need to similarly expose these parameters there. Is there a single place where they are added or is there multiple places?

These changes should be propagated to the public docs, but this PR doesn't add them new.

## Checklist:

Here's example output from:

input
Input
enhanced
After the laplacian/gaussian filters
2d_filter
After 2d filter
3d_filter
After 3d filter
type
Colored based on cell, needs to split, and artifact
type_split
Same as above except the split cells have a sphere colored in

### DIFF
--- a/benchmarks/filter_debug.py
+++ b/benchmarks/filter_debug.py
@@ -80,8 +80,14 @@ def setup_filter(
     split_ball_z_size: int = 3,
     split_ball_overlap_fraction: float = 0.8,
     n_splitting_iter: int = 10,
+    start_plane: int = 0,
+    end_plane: int = 0,
 ):
-    signal_array = read_with_dask(str(signal_path))[:10, :, :]
+    signal_array = read_with_dask(str(signal_path))
+    if end_plane <= 0:
+        end_plane = len(signal_array)
+    signal_array = signal_array[start_plane:end_plane, :, :]
+
     signal_array = np.asarray(signal_array).astype(dtype)
     shape = signal_array.shape
 
@@ -216,7 +222,7 @@ def dump_structures(
     arr = np.linalg.norm(sphere - position, axis=0)
     sphere_mask = arr <= dia / 2
 
-    with open(output_root / "structures.csv", "w") as fh:
+    with open(output_root / "structures.csv", "w", newline="") as fh:
         writer = csv.writer(fh, delimiter=",")
         writer.writerow(["id", "x", "y", "z", "volume", "volume_type"])
 
@@ -345,7 +351,7 @@ def run_filter(
             save_tiffs(
                 output_root,
                 "filtered_3d",
-                i + ball_filter.first_valid_plane,
+                n_3d_planes + ball_filter.first_valid_plane,
                 middle_planes,
                 n,
             )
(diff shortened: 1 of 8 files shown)


matham:
So what do you think about the global/local thresholding approach? I can convert this PR into that and keep the splitting parameters out of napari, unless splitting is useful in some applications and you'd like to expose it in napari and then I can keep the PR as is and make a new PR with the threshold stuff?

As I described above, the global threshold would be used to find the darkest cells, and potentially can even do what the bright tiles algorithm is meant to do. Since that can be used to eliminate any voxels that are below that threshold, which regions outside the brain are. And the local tiled threshold would help further define cells in bright regions. You'd need one, potentially two new parameters; the tile size and a second threshold value (I'm not sure if that would be needed, but perhaps tuning the tiled threshold would be helpful). And to keep the old behavior you'd just use the global one by setting tile size to zero.

I'm not suggesting to drop the bright tiling algorithm that determines in/out of brain areas currently in use, but this could be an alternate way of doing the same thing, of exuding cell looking things outside the brain.

alessandrofelder (commented):
Thanks @matham - excited about these improvements.

I am finding it hard to wrap my head around all these at once, so I've opened #535 to help me structure my thoughts.

I've made some initial tiny comments while reading the code for the first time in more depth.

> So what do you think about the global/local thresholding approach?

Here I was wondering whether the global/local thresholding could be done as a preprocessing step? 

Generally, we are happy with changes as long as we're sure `cellfinder` still works well on original application data.

matham:
This PR as is is differently doing too much at once, mainly because I started using the branch to stage all the related stuff so I can test it more easily. So I will break it off into 3+ PRs?

1. Make the docs consistent and add the split parameters everywhere (mostly `main` functions), *except* in napari so they're not exposed to end users there.
2. Add  `benchmarks/filter_debug.py` for testing the filters.
3. Expose pin memory and batch size with good defaults everywhere and cos. But also expose them in napari.
4. Make WIP PR for the local threshold changes.

This seem ok? I'd love to combine 1 and 3 because 3 will git conflict with 1 since they both touch similar lines of code. But I can also brake the branch of 3 off 1 so when 1 is merged it should be ok with 3 in terms of git?

alessandrofelder:
Sounds like a good plan!

> But I can also brake the branch of 3 off 1 so when 1 is merged it should be ok with 3 in terms of git?

Yep - 1 should be straightforward to review, and I can prioritise it, so we avoid conflicts.

matham:
See https://github.com/brainglobe/cellfinder/pull/542, https://github.com/brainglobe/cellfinder/pull/544 and https://github.com/brainglobe/cellfinder/pull/545.

PR #544, opened 2025-06-07
Visualize 2d and 3d filtering steps

matham:
## Description

This PR depends on https://github.com/brainglobe/cellfinder/pull/542.

Added a file under benchmarks (`filter_debug.py`) that given an input folder containing a list of tiff, it loads it using dask and does full 2d and 3d filtering, cell detection and cluster splitting. For each filtering step it outputs the filtered image so you can see what the step produces and tune parameters. 

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

Because it's hard to tune all these parameters when you only have the final output (candidate cells), which takes a bit to run and you have to guess at how these input parameters actually affect the outcome. There's already something like this for saving the output of the 3d filtering step, but having the 2d step output in particular is helpful. 

You can see how I used this in https://github.com/brainglobe/cellfinder/pull/532 and in https://github.com/brainglobe/cellfinder/pull/545.

**What does this PR do?**

Add the `filter_debug.py` file under `benchmarks`. If we had an `examples` folder I would have used that.

I also had to add copying settings during splitting because we change the input settings during splitting. Previously it wasn't a problem because it's each cluster was split in its own process so each settings was copied anyway. But now I was doing it all in one thread so the original settings instance was modified for future clusters.

It would be nice to be able to visualize all these debug outputs directly in napari, but that would require a lot more work and possibly not worth the complexity. Although I do wonder how most users would approach this tuning stage without being able to see how the parameters change the output. And I suspect something like this would be invaluable when tuning parameters for new samples.

## References

https://github.com/brainglobe/cellfinder/pull/532 and in https://github.com/brainglobe/cellfinder/pull/545.

I tested it with data.

## Is this a breaking change?

No.

No.

## Checklist:

### DIFF
--- a/benchmarks/filter_debug.py
+++ b/benchmarks/filter_debug.py
@@ -1,0 +1,444 @@
+"""
+Given an input folder containing a list of tiff, it loads it using dask
+and does full 2d and 3d filtering, cell detection and cluster splitting.
+For each filtering step it outputs the filtered image so you can see what
+the step produces and tune parameters.
+
+It outputs the following folders in the provided output directory (comment
+out `save_tiffs` or filtering steps if you don't want wish to run / save them):
+
+- input: The original input tiffs
+- clipped: The first step of 2d filtering where the data is clipped to a max
+      value (typically should be unchanged).
+- enhanced: After it was run through the 2d filters.
+- inside_brain: Tiled planes, with each tile indicating whether the tile is
+      inside or outside the brain.
+- filtered_2d: After the enhanced images was thresholded to a binary foreground
+      / background.
+- filtered_3d: After the 3d ball filtering.
+- struct_id: The output from the cell detector, where each voxel has an ID with
+      its cell number or zero if it's background.
+- struct_type: Voxel values are 0 for background, 1 if it's a potential cell,
+      2 if it's a structure to be split, and 3 if it's too big to be split.
+- struct_type_split: Same as struct_type except we put a sphere with value 1
+      centered on the structures that were split into cells.
+
+There's also the following files:
+
+- `structures.csv`. CSV that lists all the detected structures, their volumes,
+  and type.
+- `cells.xml`. All the detected cells before splitting structures.
+- `structs_needs_split.xml`. All the detected structures that were too big
+  to be cells that will be split.
+- `structs_too_big.xml`. All the detected structures that were too big to even
+  be split.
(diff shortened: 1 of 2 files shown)


benchmarks/filter_debug.py
```diff
@@ -0,0 +1,416 @@
...
+      / background.
+- filtered_3d: After the 3d ball filtering.
+- struct_id: The output from the cell detector, where each voxel has an ID with
+      its cell number or zero if it's background.
+- struct_type: Voxel values are 0 for background, 1 if it's a potential cell,
+      2 if it's a structure to be split, and 3 if it's too big to be split.
+- struct_type_split: Same as struct_type except we put a sphere with value 1
+      centered on the structures that were split into cells instead of `2`.
+
+There's also a `structures.csv` output that lists all the detected structures,
```
alessandrofelder, line 26: Could we also [save cell candidates in brainglobe format?
](https://github.com/brainglobe/brainglobe-utils/blob/c681d3e20ef922b5a3b051961d4443b81b28b48e/brainglobe_utils/IO/cells.py#L201)That would be incredibly helpful for documentation screenshots, and comparing versions of the code visually?

matham:
I rebased/squashed and added the xml output.

There's a bit of idiosyncrasy I think in how the cells are typed and loaded in napari. The options seems to be either artifact/unknown (what's the difference? and why when saving does artifact get turned into unknown? it's not very clear from the docs) and cell. But what if I wanted to save them as different types? E.g. here I'd have loved to be able to save them as 4 types: cell, cell after splitting, struct to be split, and struct too large to be split. But I had to fit it into those 2 types so I just ended up with 4 files.

Related, typically when saving a layer from napari, a layer of points, the xml files is named based on the layer name, which can be anything. But when I open the xml in napari, the original filename is gone and it's named as cells and not cells. This makes it hard to keep track of what layer came from which file. Plus, could there not be more than 2 cell types?

adamltyson:
Basically, the current xml format for saving cells is a historical artefact. I've been wanting to come up with a better format for a while to address these issues (and more), but not got round to it yet. In principle it shouldn't be too complex to come up with a new, much more flexible format.

matham:
I rebased on https://github.com/brainglobe/cellfinder/pull/542.

alessandrofelder (approved):

PR #546, opened 2025-06-08
Expose pin_memory to main and napari

matham:
## Description

This depends on https://github.com/brainglobe/cellfinder/pull/542.

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

`pin_memory` is already used in detection, and will be used in classification. It should be exposed so users can use it to speed up GPU data uploading.

I didn't initially add it to classification as well. But I added it because we will need it and we will need to add it to brainglobe-workflows to pass it on. And making workflows being able to pass this parameter to both detection and classification, even before we start using it, will reduce compatibility issues.

**What does this PR do?**

It exposes it in napari and in the main functions. But it defaults to not being enabled due to the requirement of the data memory not being paged out of RAM if it's used. So users will need to turn it ON manually.

## References

https://github.com/brainglobe/cellfinder/issues/540

Checked that it looks ok in napari.

## Is this a breaking change?

No. It does change the default for detection from pinning to not pinning. But I added the default when we moved detection to pytorch.

No.

## Checklist:

### DIFF
--- a/cellfinder/core/classify/classify.py
+++ b/cellfinder/core/classify/classify.py
@@ -29,6 +29,7 @@ def main(
     model_weights: Optional[os.PathLike],
     network_depth: depth_type,
     max_workers: int = 3,
+    pin_memory: bool = False,
     *,
     callback: Optional[Callable[[int], None]] = None,
 ) -> List[Cell]:
@@ -74,6 +75,12 @@ def main(
     max_workers: int
         The number of sub-processes to use for data loading / processing.
         Defaults to 8.
+    pin_memory: bool
+        Pins data to be sent to the GPU to the CPU memory. This allows faster
+        GPU data speeds, but can only be used if the data used by the GPU can
+        stay in the CPU RAM while the GPU uses it. I.e. there's enough RAM.
+        Otherwise, if there's a risk of the RAM being paged, it shouldn't be
+        used. Defaults to False.
     callback : Callable[int], optional
         A callback function that is called during classification. Called with
         the batch number once that batch has been classified.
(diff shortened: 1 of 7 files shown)


matham:
To the best of my understanding on how this works ([this is a good overview](https://docs.pytorch.org/tutorials/intermediate/pinmem_nonblock.html#memory-management-basics)). When we set `pin_memory` to `True`, it prevents the system from swapping out that memory. 

So if you try to create a `pin_memory`array larger than available pure RAM, you'd probably get a memory error because we can't use paged memory for it. If you have an array with `pin_memory` taking up a lot of RAM and you need more RAM for other purposes, more than is available in pure RAM, the `pin_memory` RAM can't be swapped out so instead the new memory would have to be paged a lot, slowing down the system.

So it effectively reduces your system RAM by the amount used by `pin_memory`. This is helpful if you have enough RAM, but the system may page out the loaded arrays for whatever reasons, slowing down GPU transactions. But if you have a ton of RAM the system probably won't page it out anyway so it won't help a lot.

E.g. on our system, with 768GB RAM, running classification on 27.6 million candidate cells, which took around 9.5 hours (with https://github.com/brainglobe/cellfinder/pull/493), there was only a couple minutes difference with pinning memory. But I expect with a system with much less RAM and perhaps slower SSDs so paging is slower, not turning ON `pin_memory` would slow down classification since the arrays may be paged out.

Side note, you don't want to pin memory in the main thread, as that may block, slowing down the pipeline. And we don't do it for detection or classification, unless you specifically set classification to not use any workers.

matham:
With a better batch size, classifying `27578643` cells took `22092s` (`6.14` hours) with pinning memory and `24397s` (`6.78` hours) without. A `38 min` difference. Which is not insignificant.

alessandrofelder:
Thanks for the detailed explanation.
Confirming I get an 99%+ match on internal large test dataset `MS_cx_left` so will merge - thanks @matham !

PR #557, opened 2025-09-21
Fix #443 by removing the leaking logs from pytest

### LINKED ISSUE #443: Tests are failing with napari 0.5.0
@alessandrofelder's hypothesis - "napari fixtures are now cleverer and fail your tests if you don't clean up your widgets properly"

We should look into this and get the tests passing again (e.g. by making sure all qt widgets have parents)

matham:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

#443 shows that some logs are leaking objects. Specifically, when fancylog was used during a test, it set the logging level of the default logger to debug without changing it back afterwards. This resulted in all subsequent tests emitting debug logs.

Next, the `in_n_out` library used by `napari` will debug log all napari objects and the pytest `caplog` fixture saves all logs forever. So when napari checks after each test whether there's a memory leak of their widgets, because caplog does leak the widgets, the tests fails.

**What does this PR do?**

This clears these logs from `caplog` after the test, if the test worked, preventing the leaks.

## References

https://github.com/brainglobe/cellfinder/issues/443, https://github.com/napari/napari/issues/8287.

Locally.

## Is this a breaking change?

No.

No.

## Checklist:

### DIFF
--- a/tests/napari/conftest.py
+++ b/tests/napari/conftest.py
@@ -1,0 +1,17 @@
+import pytest
+
+
+@pytest.fixture
+def create_napari_viewer(make_napari_viewer, caplog):
+    """Shim for make_napari_viewer, that cleans up caplog from holding
+    onto widgets. See issue #443.
+    """
+    yield make_napari_viewer
+
+    # we only need to remove the logs if test didn't fail. Napari doesn't
+    # check about memory leaks for failed tests
+    for when in ("setup", "call", "teardown"):
+        records = caplog.get_records(when)
+        indices = [i for i, r in enumerate(records) if r.name == "in_n_out"]
+        for i in indices[::-1]:
+            del records[i]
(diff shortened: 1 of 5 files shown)


IgorTatarnikov:
This looks great! I've tested it locally and it works. We have the same issue in other repos that use `napari`. I wonder if this fixture could be moved to `brainglobe-utils` so that it can be used more broadly?

matham:
That makes sense. Do you know offhand how pytest finds fixtures from other projects?

I don't see anywhere we're directly importing `make_napari_viewer` yet pytest finds it. I wonder if it'd be the same if we declare a fixture in utils and want to use it in cellfinder. I can't imagine it'd work like that. Perhaps napari doesn't something special to make it available?

PR #564, opened 2025-10-07
Pin minimum fancylog version

adamltyson:
Following #561 I realised we should pin the minimum fancylog version, otherwise just updating cellfinder will cause issues.


IgorTatarnikov (approved):
Good catch!

Issue #548, opened 2025-06-10
Napari saving training data shows done before actually done


matham:
**Describe the bug**

In Napari, when curating data, when you click "save training data" to start exporting the cubes a popup pops up right away saying "done". Even though it only actually started saving the data.

This is very confusing because it takes a second for the progress bar to start going so I thought it was actually done right away. But the folder was still empty so I deleted it and got errors of course because it was still saving.

**To Reproduce**
Hit save training data and you'll see the popup appear.

**Expected behaviour**
For there to be no popup or to show it after it's actually done exporting all files.

**Computer used (please complete the following information):**

- OS: Windows
- Version 11

adamltyson:
Changing it so that the "done" doesn't pop up until after it's confirmed that the training data is saved seems like a good idea

PR #588, opened 2026-02-12
Adds support for normalizing cuboid data before classifiying

matham:
## Description

This PR depends on https://github.com/brainglobe/cellfinder/pull/493. Only the last commit in this PR is unique to this PR.

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

It's fairly standard when distributing pre-trained models, in particular image networks, to indicate how the input should be normalized, if at all. E.g. `mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]` has been [used for ImageNet based models](https://docs.pytorch.org/tutorials/beginner/transfer_learning_tutorial.html#load-data) because that was the statistics of the ImageNet data. The reason is that the model will be more accurate if your data's statistics/input range matches the pre-trained dataset's statistics.

The pre-trained model distributed by brainglobe was trained on cells with implicit intensity statistics. So it would be nice if when using on other datasets we can normalize our own data to match that distribution. But it is not specified. Additionally, input data can vary wildly in intensity between channels and labeling types and so if you do transfer learning and want to provide a minimal set of new cells to train on, if you happen to fine tune on cells from images where the overall image is bright, trying to use it on images that are darker may result in worse classification.

When training on my own cells, I found that adding such normalization helped improve accuracy as you can see in the images below.

**What does this PR do?**

This PR add the option to normalize the training data, the classification data, and to save the image statistics into the yaml files when curating training data. Specifically,

1. During curation, it'll sample the z-stack of each channel and compute the mean/std and dump that into the "curated" yaml file. E.g. here's how the yaml file will look:
   ```py
   data:
   - bg_channel: 1
     cell_def: ''
     cube_dir: tests/data/integration/training/cells
     signal_channel: 0
     type: cell
     signal_mean: 241.31
     signal_std: 154.92
     bg_mean: 650.94
     bg_std: 217.90
   ```
   It adds the parameters `normalize_channels: bool = False` and `normalization_down_sampling: int = 32` to control this.
2. During training, we load these yaml files, and if selected, we normalize the cuboids by the mean/std of the dataset from which the cuboids originated. After training, the model expect input cuboids to have a mean of 1 and std of 1. This is controlled by the parameter `normalize_channels: bool = False`.
3. During classification, we similarly sample each channel in the z-stack and normalize the cuboids that want to classify by the mean/std of the corresponding channel. This is similarly controlled by the parameters `normalize_channels: bool = False` and `normalization_down_sampling: int = 32`.

By default all the normalization is turned OFF.

## References

https://github.com/brainglobe/cellfinder/pull/493.

I tested this on my own curated data that was continued training from the pre-trained brainglobe models. As I was getting cellfinder classification to work on our c-fos and fostrap images I found the model wasn't performing good enough. So I added normalization which improved the model.

I re-did those experiments systematically on our training data just to illustrate the benefits. It seems that normalization is most helpful when you have less training data and when you're using a smaller learning rate. Both makes sense intuitively. For the former it's because if you don't have a lot of training data and they all have different statistics, the model has a harder time converging. For the same reason, a lower learning rate will likely get more easily stuck in local minima.

Here's the plotted train and test accuracy and loss for training on my data:

There are 4 plots, illustrating two different learning rates (`LR` = `1e-3` and `1e-4`), and two different values for the size of the training data (`f` = `10%` and `50%`, meaning the fraction of the training data used for testing - this is a quick way to reduce training data size by using `50%` for testing). `N=n` indicates the number of lines in the plot, which indicates the number of independent replications of the training. FYI, each replication uses a different train/test split, which explains the distribution of lines for a given condition.

But overall you can see how normalizing improves model performance, and this is only using training data from similar images (data augmentation was `25%` using `resnet50_tv` using a batch size of `32` and with `--continue-training`).

train_accuracy
test_accuracy
train_loss
test_loss

On the other hand, I tried to estimate the serial2p data's statistics from the provided cubes (already a problem because we want the statistics of the images they came from, not the extracted cubes) and saw little benefit when re-training the brainglobe model. I think this is because there's so much data in the training set that it really doesn't matter. Or the serial2p data are easier/cleaner to classify.

## Is this a breaking change?

No.

There are new parameters. But, I think it may be worth to release a new pre-trained brainglobe model trained on normalized inputs.

## Checklist:

### DIFF
--- a/cellfinder/napari/curation.py
+++ b/cellfinder/napari/curation.py
@@ -663,9 +663,11 @@ class CurationWidget(QWidget):
         self.non_cells_to_extract = list(set(self.non_cells_to_extract))
 
     def _calculate_channel_stats(self):
+        self.update_status_label("Estimating signal mean/std...")
         signal_stat = dataset_mean_std(
             self.signal_layer.data, self.normalization_n_sampling_planes
         )
+        self.update_status_label("Estimating background mean/std...")
         bg_stat = dataset_mean_std(
             self.background_layer.data, self.normalization_n_sampling_planes
         )
--- a/cellfinder/napari/curation.py
+++ b/cellfinder/napari/curation.py
@@ -17,6 +17,7 @@ from qt_niu.interaction import (
     add_button,
     add_combobox,
     add_float_box,
+    add_int_box,
 )
 from qtpy import QtCore
 from qtpy.QtWidgets import (
@@ -32,6 +33,7 @@ from cellfinder.core.classify.cube_generator import (
     CuboidArrayDataset,
     CuboidBatchSampler,
 )
+from cellfinder.core.tools.image_processing import dataset_mean_std
 
 # Constants used throughout
 WINDOW_HEIGHT = 750
@@ -68,6 +70,7 @@ class CurationWidget(QWidget):
         self.save_empty_cubes = save_empty_cubes
         self.max_ram = max_ram
         self.voxel_sizes = [5, 2, 2]
+        self.normalization_down_sampling = 32
         self.batch_size = 64
         self.viewer = viewer
 
@@ -176,6 +179,9 @@ class CurationWidget(QWidget):
     def _set_voxel_size(self, value: float, index: int) -> None:
         self.voxel_sizes[index] = value
 
+    def _set_normalization_down_sampling(self, value: int) -> None:
+        self.normalization_down_sampling = value
+
     def add_loading_panel(self, row: int, column: int = 0):
         self.load_data_panel = QGroupBox("Load data")
         self.load_data_layout = QGridLayout()
@@ -233,32 +239,46 @@ class CurationWidget(QWidget):
         )
         box_x.valueChanged.connect(partial(self._set_voxel_size, index=2))
         self.voxel_sizes_boxes = box_z, box_y, box_x
+
+        box_norm = add_int_box(
+            self.load_data_layout,
+            self.normalization_down_sampling,
+            1,
+            1000,
+            "Normalization down-sampling",
+            6,
+            tooltip="Down-sampling factor of the z-dimension used to calculate"
+            " the mean and std of the dataset. Used to normalize the "
+            "channels during training.",
+        )
+        box_norm.valueChanged.connect(self._set_normalization_down_sampling)
+        self.norm_sampling_box = box_norm
         self.training_data_cell_choice, _ = add_combobox(
             self.load_data_layout,
             "Training data (cells)",
             self.point_layer_names,
-            6,
+            7,
             callback=self.set_training_data_cell,
         )
         self.training_data_non_cell_choice, _ = add_combobox(
             self.load_data_layout,
             "Training_data (non_cells)",
             self.point_layer_names,
-            row=7,
+            row=8,
             callback=self.set_training_data_non_cell,
         )
         self.mark_as_cell_button = add_button(
             "Mark as cell(s)",
             self.load_data_layout,
             self.mark_as_cell,
-            row=8,
+            row=9,
             tooltip="Mark all selected points as non cell. Shortcut: 'c'",
         )
         self.mark_as_non_cell_button = add_button(
             "Mark as non cell(s)",
             self.load_data_layout,
             self.mark_as_non_cell,
-            row=8,
+            row=9,
             column=1,
             tooltip="Mark all selected points as non cell. Shortcut: 'x'",
         )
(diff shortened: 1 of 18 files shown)


cellfinder/napari/curation.py
```diff
@@ -632,21 +652,39 @@ def convert_layers_to_cells(self):
...
+        signal_stat = dataset_mean_std(
+            self.signal_layer.data, self.normalization_down_sampling
+        )
+        bg_stat = dataset_mean_std(
+            self.background_layer.data, self.normalization_down_sampling
+        )
+        return signal_stat, bg_stat
+
     def __save_yaml_file(self):
+        signal_stat, bg_stat = self._calculate_channel_stats()
```
alessandrofelder, line 677: Quick check that I have understood this correctly: we will from now on always save the per-channel stats after curation (and by default, we ignore them during training for backwards compatibility)? Presumably that won't affect the speed at which we save files much?

alessandrofelder (approved):
Thanks a lot for this @matham 

I have finally been able to review this 😅 

I confirm this is backwards compatible both in result and performance for our viral tracing `MS_cx_left` data. Noting results for posterity here:
```
Successfully installed cellfinder-1.10.1.dev1+ge127361a0
Running cellfinder from the main branch...
[...
Finished. Total time taken: 0:29:24.593917                            main.py:95

Successfully installed cellfinder-1.9.1.dev5+g59db9a492
Running cellfinder from the PR branch...
[...]
Finished. Total time taken: 0:29:23.058641                            main.py:95

Comparing the results...
Number of cells only found with cellfinder from main: 0
Number of cells only found with cellfinder from 588: 0
Number of matched cells: 70411
Matching proportion: 1.0
```
On the code changes side, all looks reasonable to me.
I just left one small question for something I wanted to double-check.

cellfinder/napari/curation.py
matham: Correct! For my data of about 2000 giant planes, it take about a minute per channel to calculate the stats.

However...due to #548, the work happens in the main thread largely blocking the GUI. So while the stats are calculated for the couple of minutes it looks like the GUI is just frozen - you have to look at the terminal to see a progress bar. Once the stats are done, the progress bar in the GUI is updated with number of cells saved, even if the GUI stays frozen. Perhaps we need to add a message initially that stats are being calculated. Or maybe this can all be fixed if we move the work to a second thread?

matham:
One consideration you may want to look at and let me know what you think. With this change, we simply sample every `n` planes. E.g. with `2000` planes down-sampled by `32`, we use `63` planes.

But for a smaller dataset, say `100` planes, the default `32` will only use `3` planes with 2 of them towards the start and end, which are probably dark planes. In this case, perhaps it would have been better if the parameter specified the number of planes we want to use in the sampling, say `10`, and from that we calculate the downs-sampling.

So e.g. if we defaulted it to `50`, then if the dataset was less than `50`, we used all planes. If it was `100`, we'd use every second plane. And for my dataset we'd use every 40th plane. Do you think this a better approach?

--------

Also, I was thinking, if possible the pre-trained model brainglobe provides should have a normalized option. This would mean the stats for the training data will have to be calculated from the original datasets and added to the training data. But this would make it possible for people to continue training with their data while using normalization, probably speeding up fine-tuning. Based on whether normalization is selected, cellfinder would download the correct weights. And ideally, long term, cellfinder would use normalization be default because there's no reason not to other than backwards compat. What do you think?

alessandrofelder:
>So e.g. if we defaulted it to 50, then if the dataset was less than 50, we used all planes. If it was 100, we'd use every second plane. And for my dataset we'd use every 40th plane. Do you think this a better approach?

Yea, great point - I think this is a better approach - would you have capacity to fix in this PR?

This has made me realise: I guess this PR means that if we load a subset of a brain instead of the whole brain, but curate exactly the same in both cases, we will get different results due to the statistics of the image stack? (I think this is fine)

alessandrofelder:
> Also, I was thinking, if possible the pre-trained model brainglobe provides should have a normalized option. This would mean the stats for the training data will have to be calculated from the original datasets and added to the training data. But this would make it possible for people to continue training with their data while using normalization, probably speeding up fine-tuning. Based on whether normalization is selected, cellfinder would download the correct weights. And ideally, long term, cellfinder would use normalization be default because there's no reason not to other than backwards compat. What do you think?

I agree, and have opened #617 to track (at least the first part) of your suggestion.

matham:
>Yea, great point - I think this is a better approach - would you have capacity to fix in this PR?
Yes, will do soonish.

>This has made me realise: I guess this PR means that if we load a subset of a brain instead of the whole brain, but curate exactly the same in both cases, we will get different results due to the statistics of the image stack? (I think this is fine)

Yes! Assuming the stats of the subset is different enough and more importantly, the model wasn't trained with enough robustness/data to account for different data, you would get different results.

matham:
I rebased and changed sampling factor to number of planes, as discussed!

cellfinder/napari/curation.py
alessandrofelder: Opened #626 to take this discussion further. Maybe a progress bar in the GUI is sufficient here.

alessandrofelder (approved):
Awesome, thanks a lot @matham

PR #592, opened 2026-02-19
Vectorize numpy operations in structure splitting

aymuos15:
## Description
                                                                                                                                                                                               
   **What is this PR**                                                                    

   - [x] Other

   Refactoring / performance improvement.

   **Why is this PR needed?**

   The post-processing in `split_cells` used element-by-element Python loops and a per-point helper function (`check_centre_in_cuboid`) instead of idiomatic vectorized numpy. This is predominantly slower and the patch is much cleaner. 

   **What does this PR do?**

   Replaces three loop/element-wise patterns in `split_cells()` with vectorized numpy equivalents:

   1. **`relative_orig_centre`** — element-by-element subtraction → `orig_centre - orig_corner`
   2. **Outlier filtering** — Python list comprehension calling `check_centre_in_cuboid` per point → `np.all(..., axis=1)` boolean mask
   3. **`absolute_centres`** — per-column addition into a pre-allocated array → `relative_centres + orig_corner`

   The helper function `check_centre_in_cuboid` is removed as its logic is now inlined.

   ### Benchmark results (median, 50 repeats)

   | n_centres | old (µs) | new (µs) | speedup |
   |----------:|----------:|----------:|--------:|
   | 10 | 32.3 | 15.3 | 2.11x |
   | 100 | 389.9 | 7.9 | 49.50x |
   | 1,000 | 1421.2 | 34.9 | 40.66x |
   | 10,000 | 14850.2 | 276.7 | 53.67x |
   | 100,000 | 149660.4 | 2817.0 | 53.13x |

   Correctness verified: old and new outputs match (`np.allclose`) at every size.

   ## References

   N/A

   - Existing integration test `test_underflow_issue_435` exercises `split_cells` end-to-end and continues to pass.
   - Standalone benchmark script verified correctness via `np.allclose` between old and new implementations across 5 input sizes.

   ## Is this a breaking change?

   No. Pure internal refactor with identical outputs.

   No.

   ## Checklist:

   - [x] The code has been formatted with [pre-commit](https://pre-commit.com/)"

### DIFF
--- a/cellfinder/core/detect/filters/volume/structure_splitting.py
+++ b/cellfinder/core/detect/filters/volume/structure_splitting.py
@@ -179,33 +179,6 @@ def iterative_ball_filter(
     return ns, centres
 
 
-def check_centre_in_cuboid(centre: np.ndarray, max_coords: np.ndarray) -> bool:
-    """
-    Checks whether a coordinate is in a cuboid.
-
-    Parameters
-    ----------
-
-    centre : np.ndarray
-        x, y, z coordinate.
-    max_coords : np.ndarray
-        Far corner of cuboid.
-
-    Returns
-    -------
-    True if within cuboid, otherwise False.
-    """
-    relative_coords = centre
-    if (relative_coords > max_coords).all():
-        logger.info(
-            'Relative coordinates "{}" exceed maximum volume '
-            'dimension of "{}"'.format(relative_coords, max_coords)
-        )
-        return False
-    else:
-        return True
-
-
 def split_cells(
     cell_points: np.ndarray, settings: DetectionSettings
 ) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


cellfinder/core/detect/filters/volume/structure_splitting.py
```diff
@@ -290,18 +257,18 @@ def split_cells(
...
         original_max_coords = np.array(original_bounding_cuboid_shape)
-        relative_centres = np.array(
-            [
-                x
-                for x in relative_centres
-                if check_centre_in_cuboid(x, original_max_coords)
-            ]
-        )
+        # keep centres where not all coordinates exceed max
+        mask = ~np.all(relative_centres > original_max_coords, axis=1)
```
alessandrofelder, line 261: I realise this is a correct reimplementation of the existing codebase but I wonder whether that was a bug? It's sufficient for the coordinate along _any_ axis to be outside the cube for it to be considered an outlier (I think). I have a vague, possibly wrong memory of @matham also noticing this somewhere.

I'd maybe also flip the logic to reduce the mental gymnastics when reading the code
```suggestion
        is_outlier = np.any(relative_centres > original_max_coords, axis=1)
```
And adapt the conditional accordingly?

alessandrofelder (changes requested):
Thanks @aymuos15 - I think this is a sensible implementation. I made a small suggestion to improve the code legibility and (possibly?) fix a bug while we are here :)

cellfinder/core/detect/filters/volume/structure_splitting.py
matham: I fixed the bug you remember in https://github.com/brainglobe/cellfinder/pull/591 ☺️ I think it may make some of these changes unneeded.

aymuos15:
Thank you for the reviews and discussion! I would be in the group of letting 591 merge and then getting back to this then? Would that be okay?

PR #593, opened 2026-02-23
Fix OR vs AND logic error in get_model

aymuos15:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

The original condition in `get_model` — `if existing_model is not None or network_depth is None` — has a logic error. When both `existing_model` and `network_depth` are `None`, the `or` evaluates to `True` (because `network_depth is None`), causing `load_model(None)` to be called, which crashes with an unhelpful Keras error.

**What does this PR do?**

- Replaces the compound `or` condition with a clear `if / elif / else` chain so that `existing_model` always takes precedence when provided
- Adds an explicit `ValueError` guard when neither `existing_model` nor `network_depth` is supplied
- Updates the docstring to clearly describe parameter precedence rules
- Adds tests for the existing-model-takes-precedence path, the depth-only build path, and the both-`None` error case

## References

None.

Three new unit tests added to `tests/core/test_unit/test_classify/test_tools.py`:

- `test_get_model_existing_takes_precedence` — verifies `load_model` is called (not `build_model`) when both arguments are provided
- `test_get_model_builds_with_depth_only` — verifies `build_model` is called when only `network_depth` is given
- `test_get_model_no_model_no_depth` — verifies `ValueError` is raised when both are `None`

All 5 tests pass locally. `get_model` now has 100% line coverage.

## Is this a breaking change?

No. The only behavioral change is for the `(None, None)` case, which was already broken (crashed with an unhelpful Keras error). It now raises a clear `ValueError` instead.

No. This is an internal bug fix with no user-facing API changes.

## Checklist:

### DIFF
--- a/cellfinder/core/classify/tools.py
+++ b/cellfinder/core/classify/tools.py
@@ -18,24 +18,33 @@ def get_model(
     inference: bool = False,
     continue_training: bool = False,
 ) -> Model:
-    """Returns the correct model based on the arguments passed
-    :param existing_model: An existing, trained model. This is returned if it
-    exists
-    :param model_weights: This file is used to set the model weights if it
-    exists
-    :param network_depth: This defines the type of model to be created if
-    necessary
-    :param learning_rate: For creating a new model
+    """Returns the correct model based on the arguments passed.
+
+    If ``existing_model`` is provided it is loaded and returned directly,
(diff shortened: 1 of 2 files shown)


alessandrofelder:
Superceded by #643

PR #594, opened 2026-02-27
Python api napari update

vijayalakshmipokala:
What this PR does:
Fixes the main() function in detect.py for Python API / Napari.
Adds test_detect_main.py to ensure correct 3D signal handling and proper error raising.
Why this PR is needed:
Resolves import errors and failing integration tests.
Testing:
test_detect_main.py added and all local tests pass.
Breaking change: No
Documentation update: Not needed
Checklist:
[x] Code tested locally
[x] Tests added
[x] Pre-commit formatting applied


vijayalakshmipokala:
Hi @adamltyson , thanks for pointing that out — that’s on me for not explaining it clearly.

This PR is focused on the Python API side used by the napari plugin. While working on the napari update, I noticed that the detect.main() entry point was a bit unclear/inconsistent in how it’s defined versus how it’s actually called.

What I’ve done here is :
Make the detect.main() definition explicit and consistent with how the Python API expects to use it
Add a small, targeted test to lock down that behavior so it doesn’t regress

There are no changes to the detection logic itself, outputs, or defaults — this is purely a cleanup/clarification of the API surface plus test coverage.

Happy to adjust the scope or split this up if that would be preferred.

adamltyson:
Hi @vijayalakshmipokala, thanks for your message. It's not clear to me that this PR improves anything, particularly because the biggest change is to delete docstrings.

I will close this PR now. If you you would like to contribute, please raise an issue outlining where the code could be improved, and let the maintainers and other community members discuss first. Thanks!

Issue #352, opened 2020-06-18
[Feature] Make background channel optional


adamltyson:
Although the background channel is useful for the cell classification and registration steps, it is not always practical to double the data collected (and in the case of some fluorophores, a channel with only auto-fluorescence can be hard to obtain). It should be possible to supply one channel and:
* Only register this one channel
* Use 3D, rather than 4D input to the classification network (at the expense of poorer performance).

IgorTatarnikov:
I've done some experiments with training a model that just used the signal channel. These were preliminary and incomplete, but they should give an idea of feasibility and the necessary next steps. All training here was done using the original data set used to train the default model and can be found [here](https://gin.g-node.org/cellfinder/training_data).

This is the loss curve with either 1-channel or 2-channels using the default 50 layer architecture:

| Loss | Accuracy |
|--------|--------|
| <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/6d404ecc-ab44-42f8-ad84-276e977014ee" /> | <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/45397e6e-f521-4330-9eec-3c5bab3eb773" /> | 

And confusion matrices on a held out data set:

| 2-channel | 1-channel |
|--------|--------|
| <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/e322cff5-6b78-4e2d-9677-8a42b19b0846" /> | <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/b2798018-1a54-4d3d-8a20-f979103b45bf" /> | 


I also tested using the smaller 14 layer architecture with 2 channels and 1 channel only:

| Loss | Accuracy |
|--------|--------|
| <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/1ee377d8-7d4f-47ba-85e2-2af50242c520" /> | <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/0c446bdc-df87-4d9a-9752-08efbb0f3c35" /> | 

And confusion matrices:

| 2-channel | 1-channel |
|--------|--------|
| <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/e756138e-7891-4ae5-a4a6-1608e1b6057c" /> | <img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/e4838ba5-ba83-4256-bb60-6a72eeba0a8f" /> |

PR #607, opened 2026-03-21
Fix #605: make background_array optional when skip_classification=True

### LINKED ISSUE #605: [BUG]: background channel is required even when skip_classification=True (single-channel flow breaks)
**Describe the bug**
I found that the Python API still requires a background channel even when `skip_classification=True`.
This seems wrong because if classification is skipped, background should not be mandatory.
Right now this blocks single-channel detection-only use cases.

**To Reproduce**
1. Create a valid 3D `signal_array`.
2. Call `cellfinder.core.main.main(...)` with:
   - `signal_array=<3D array>`
   - `voxel_sizes=(5, 2, 2)` (any valid voxel sizes)
   - `skip_classification=True`
   - do not pass `background_array`
3. It fails with:
   - `TypeError: main() missing 1 required positional argument: 'background_array'`

I also checked classification function behavior with `background_array=None` and saw:
- `AttributeError: 'NoneType' object has no attribute 'ndim'`

**Expected behaviour**
If `skip_classification=True`, run should work with signal-only input.
Background should only be required when classification is enabled.

**Log file**
No separate log file attached yet.
I reproduced using local script run:
`python -u _bug_demo.py`
and got the errors above consistently.

**Screenshots**
No screenshots (this report is based on Python/API repro, not UI testing).

- OS: Windows
- Version: Windows 11

**Additional context**
This seems related to feature request #352 (make background channel optional).
I think this is a concrete blocker bug inside that larger feature work.

KpradeepKumar25:
Background_array was always required even when skip_classification=True. This fix makes it default to None and adds
ValueError with a clear  message when classification is attempted without a background array.

Fixes #605

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

background_array was always a required argument even when skip_classification=True. This blocked researchers who only have 
a single channel (signal only) from using cellfinder for detection  without classification.

**What does this PR do?**

- Makes background_array default to None instead of being required
- Adds a clear ValueError message if someone tries to run 
  classification without providing background_array
- Updates the docstring to reflect that background_array is optional

## References

Fixes #605
Related to #352

Manually verified that calling main() with skip_classification=True and no background_array no longer raises a TypeError.
Also verified that calling main() with skip_classification=False and no background_array now raises a clear ValueError.

## Is this a breaking change?
No. Users who already pass background_array will see no change. Users who pass skip_classification=True without background_array  will now get the expected behaviour instead of a TypeError.

The docstring for background_array in main() has been updated  to reflect it is now optional.

## Checklist:

### DIFF
--- a/cellfinder/core/main.py
+++ b/cellfinder/core/main.py
@@ -10,8 +10,8 @@ from cellfinder.core.train.train_yaml import depth_type
 
 def main(
     signal_array: types.array,
-    background_array: Optional[types.array] = None,
     voxel_sizes: Tuple[float, float, float],
+    background_array: Optional[types.array] = None,
     start_plane: int = 0,
     end_plane: int = -1,
     trained_model: Optional[os.PathLike] = None,
@@ -55,10 +55,11 @@ def main(
     ----------
     signal_array : numpy.ndarray or dask array
         3D array representing the signal data in z, y, x order.
-    background_array : numpy.ndarray or dask array
-        3D array representing the signal data in z, y, x order.
     voxel_sizes : 3-tuple of floats
         Size of your voxels in the z, y, and x dimensions (microns).
+    background_array : numpy.ndarray or dask array, optional
+        3D array representing the background data in z, y, x order.
+        Required when skip_classification=False. Defaults to None.
     start_plane : int
         First plane index to process (inclusive, to process a subset of the
         data).
@@ -226,7 +227,8 @@ def main(
     if not skip_classification:
         if background_array is None:
             raise ValueError(
-                "background_array is required when skip_classification=False"
+                "background_array is required when "
+                "skip_classification=False"
             )
         install_path = None
         model_weights = prep.prep_model_weights(
--- a/cellfinder/core/main.py
+++ b/cellfinder/core/main.py
@@ -10,7 +10,7 @@ from cellfinder.core.train.train_yaml import depth_type
 
 def main(
     signal_array: types.array,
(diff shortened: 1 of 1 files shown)


alessandrofelder (changes requested):
Thanks for contributing @KpradeepKumar25 

I think it's unfortunately too confusing/not backwards-compatible to switch the order of the positional arguments as this PR suggests. Many workflows that depend on this will need to switch the order of the arguments with which they call the `main` function if we merged this.

I suggest instead that we keep the order of the arguments, but allow `None` to be explicitly passed for the background array, rather than being the default.

Also, this PR should add a test that checks that the expected error is raised when `background_array=None` is passed.

[Edit] I would be grateful if you could also resolve the merge conflicts.

KpradeepKumar25:
Thank you for the review @alessandrofelder  I appreciate the clear feedback. I understand the concern about breaking existing workflows by changing the argument order — I will revert it back to the original position and instead allow None to be passed explicitly for background_array.
I will also add a pytest test to verify the ValueError is raised correctly, and I will resolve the merge conflicts.
I want to be transparent — I am currently in my final semester exams, so I am taking a bit of time to study the codebase more carefully before updating the PR. I want to make sure I understand the full picture properly rather than rushing a fix. I will update the PR as soon as my exams are done. Thank you for your patience!

Issue #337, opened 2022-02-01
Input validation


alessandrofelder:
Given that cellfinder can take a while to run, we should check for valid inputs before running the computationally expensive parts.

Currently, we can e.g. input to keep more CPUs free than exist on the local machine (not sure what happens, maybe it uses 1?) and we can input an invalid model and it runs through the entire cell detection process before failing inside keras.

It might be worth splitting the input parameters into structs (e.g. cell detection parameters, cell classification parameters, ...). Each with their validator. See also brainglobe/cellfinder-napari#41 .

The validation should be tested (negative tests to check error is thrown for invalid inputs).
And a the error message displayed in a widget on-screen for feedback to the user.

PR #608, opened 2026-03-26
Validate model inputs before running detection

iliasmahboub:
If classification is enabled but the model or weights path is wrong,
cellfinder runs the entire detection before hitting the error. This
moves the validation to happen first so it fails fast.

Addresses #337

`prep_model_weights` now runs before detection starts. Also added a
check in `prep_models` for when a user passes a custom weights path
that doesn't exist.

Run `python -m pytest tests/core/test_unit/test_main.py -v`

### DIFF
--- a/cellfinder/core/main.py
+++ b/cellfinder/core/main.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 from typing import Callable, List, Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


alessandrofelder (approved):
Thanks @iliasmahboub!

I appreciate how you've kept the PR short and focused.
All looks good to me.

PR #610, opened 2026-03-29
fix: allow detection-only runs without background channel

### LINKED ISSUE #605: [BUG]: background channel is required even when skip_classification=True (single-channel flow breaks)

NiShITa-code:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

`cellfinder.core.main.main(...)` currently requires a `background_array` even when `skip_classification=True`.

That blocks single-channel detection-only use cases and causes the Python API to fail before the skip-classification logic is reached. This is also a concrete blocker related to optional-background support discussed in #352.

**What does this PR do?**

- makes `background_array` optional at the `cellfinder.core.main.main(...)` entrypoint
- allows detection-only runs when `skip_classification=True` and no background channel is provided
- preserves the requirement for `background_array` when classification is enabled
- adds regression tests for both the allowed detection-only case and the classification-required error case

## References

- Closes #605
- Related to #352

I added regression tests covering:
- detection-only execution without `background_array`
- raising a clear error when classification is requested without `background_array`

I also manually reviewed the changed code paths to ensure classification still requires background input.

## Is this a breaking change?

No.

This change preserves existing behavior for classification workflows and only relaxes the API for detection-only runs when `skip_classification=True`.

No documentation changes were made in this PR.

This is a small bug fix to the Python API behavior. If preferred, I’m happy to add a short note to the API docs in a follow-up PR.

## Checklist:

### DIFF
--- a/cellfinder/core/main.py
+++ b/cellfinder/core/main.py
@@ -10,8 +10,8 @@ from cellfinder.core.train.train_yaml import depth_type
 
 def main(
     signal_array: types.array,
-    background_array: types.array,
-    voxel_sizes: Tuple[float, float, float],
+    background_array: types.array | Tuple[float, float, float] | None = None,
+    voxel_sizes: Tuple[float, float, float] | None = None,
     start_plane: int = 0,
     end_plane: int = -1,
     trained_model: Optional[os.PathLike] = None,
@@ -55,8 +55,10 @@ def main(
     ----------
     signal_array : numpy.ndarray or dask array
         3D array representing the signal data in z, y, x order.
-    background_array : numpy.ndarray or dask array
-        3D array representing the signal data in z, y, x order.
+    background_array : numpy.ndarray or dask array, optional
+        3D array representing the background data in z, y, x order. This is
+        required when classification is enabled and optional when
+        ``skip_classification=True``.
     voxel_sizes : 3-tuple of floats
         Size of your voxels in the z, y, and x dimensions (microns).
     start_plane : int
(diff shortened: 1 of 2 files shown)


cellfinder/core/main.py
```diff
@@ -188,6 +190,21 @@ def main(
     from cellfinder.core.detect import detect
     from cellfinder.core.tools import prep
 
+    # Preserve the historical positional API while allowing callers to omit
+    # background data for detection-only runs.
+    if voxel_sizes is None:
```
alessandrofelder, line 195: I don't think this is a clean way to retain the API, because while the positional API is maintained, you could pass the voxel sizes via the background array keyword argument, and these are very different things!

alessandrofelder (changes requested):
Thanks for contributing @NiShITa-code and sorry it took so long to get back to you.

I understand your approach, but I don't think it's clean enough - have explained in a comment.

I suggest instead that we keep the arguments, but allow None to be explicitly passed for the background array, rather than being the default.

PR #621, opened 2026-06-01
Show "Done" only after training data is actually saved.

### LINKED ISSUE #548: Napari saving training data shows done before actually done

aymuos15:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

In the napari curation widget, clicking "Save training data" showed a "Done" popup immediately, before any cubes had been written to disk. The output folder appeared empty, so users assumed the export had finished and could delete the folder mid-write, causing errors. The async (non-blocking) save path launched the cube-extraction `thread_worker` and then synchronously wrote `training.yaml` and showed "Done", without waiting for the worker to finish.

**What does this PR do?**

It moves the YAML write, the "Done" notification, and the status-label reset into a single `__finish_save` helper, invoked synchronously for the `block=True` path and from the worker's `returned` signal for the async path, so "Done" appears only after all cubes are saved.

## References

Fixes #548.

Added a regression test `test_async_save_done_only_after_cubes_written` in `tests/napari/test_curation.py` that drives the async path and asserts no files exist and "Done" has not been shown immediately after the call returns, then waits for the worker (`qtbot.waitUntil`) and confirms "Done" fires exactly once with two cell and two non-cell cubes written. 

## Is this a breaking change?

No.

No.

## Checklist:

### DIFF
--- a/cellfinder/napari/curation.py
+++ b/cellfinder/napari/curation.py
@@ -469,10 +469,8 @@ class CurationWidget(QWidget):
             if self.output_directory is not None:
                 self.__prep_directories_for_save()
                 self.__extract_cubes(block=block)
-                self.__save_yaml_file()
-                show_info("Done")
-
-            self.update_status_label("Ready")
+                if block:
+                    self.__finish_save()
 
     def __prep_directories_for_save(self):
         self.yaml_filename = self.output_directory / "training.yaml"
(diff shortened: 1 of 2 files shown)


alessandrofelder (approved):
This is very nice - thanks a lot @aymuos15 .

I have tested manually locally too and it works as expected and very smoothly :) 

I don't think there is a good reason to start our functions here with a dunder (one underscore should be enough?) but I assume you've named the new function `__finish_save` for consistency with the others? (this should maybe be an issue.)

aymuos15:
Thank you very much!

I'll have a look into the naming angle and see if it warrants an issue.

PR #631, opened 2026-06-21
Fix test that failed due to reliance on randomness

matham:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

To test that cube normalization worked, I had a test that created a volume using a normal distribution and then I tested that after normalization the mean was zero and std was 1, within some bounds. The problem is that this test failed in this unrelated PR https://github.com/brainglobe/cellfinder/actions/runs/27896437125/job/82548769409?pr=628. That's because when taking cuboids of the volume, their mean/std was out of the testing bounds simply due to randomness leading to test failure.

**What does this PR do?**

This creates the volume without relying on randomness, making it so that every cuboid should have similar mean/std. We do it by alternating the values around the mean by +/- std.

## References

https://github.com/brainglobe/cellfinder/actions/runs/27896437125/job/82548769409?pr=628

The test works.

## Is this a breaking change?

No.

No.

## Checklist:

### DIFF
--- a/tests/core/test_unit/test_classify/test_cube_gen.py
+++ b/tests/core/test_unit/test_classify/test_cube_gen.py
@@ -1257,8 +1257,18 @@ def _get_volume_with_stats(normalize):
         back_norm = back_mean, back_std
 
     volume = np.empty((20, 20, 30, 2), dtype=np.float32)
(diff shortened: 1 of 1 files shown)


alessandrofelder (approved):
Thanks @matham - I just had a test spuriously fail for this reason on a different PR which motivated me to get this merged sooner 😂

PR #633, opened 2026-06-24
Validate test-uv action in CI

AlgoFoe:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other

**Why is this PR needed?**
- To validate the experimental ```test-uv``` action.

**What does this PR do?**
- Replaces the current tox-based test action with the experimental uv-based test action.

## References
- [PR -164](https://github.com/neuroinformatics-unit/actions/pull/164) 
- [Issue -93](https://github.com/brainglobe/BrainGlobe/issues/93)

- Tested locally with:
   - ```uv pip install -e .[dev]```

   - ```uv run pytest```

## Is this a breaking change?
- No.

- No.

## Checklist:


.github/workflows/test_and_deploy.yml
```diff
@@ -192,35 +200,34 @@ jobs:
...
       - name: Checkout brainglobe-workflows
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           repository: 'brainglobe/brainglobe-workflows'
 
-      - name: Set up Python 3.12
-        uses: actions/setup-python@v3
+      - name: Install uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39
```
alessandrofelder: Nitpicky: Does this need a comment with the version next to it?

.github/workflows/test_and_deploy.yml
```diff
@@ -89,21 +89,26 @@ jobs:
         uses: tlambert03/setup-qt-libs@v1
       # Setup VTK with headless display
       - uses: pyvista/setup-headless-display-action@v3
-      # Run all tests
-      - uses: neuroinformatics-unit/actions/test@v2
+
+      # dev,napari replaces tox's "extras = dev\n        napari"
```
alessandrofelder:
```suggestion
```
I would leave out this comment because it requires context of the migration from tox to uv, which we might not remember in the future - so it could become confusing! (also what the code does is well-expressed in the `install-extras: 'dev,napari'` line so a comment is not needed)

.github/workflows/test_and_deploy.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           key: ${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/pooch_registry.txt') }}
       # Cache the tensorflow model so we don't have to remake it every time
       - name: Cache brainglobe directory
-        uses: actions/cache@v3
+        uses: actions/cache@v4
```
alessandrofelder: Shall we take the opportunity to switch to SHA for this and other cache actions? Or would you prefer doing this in a separate PR?

.github/workflows/test_and_deploy.yml
```diff
@@ -89,21 +89,26 @@ jobs:
...
         if: github.event_name == 'schedule'
-        uses: neuroinformatics-unit/actions/test@v2
-        with:
-          python-version: ${{ matrix.python-version }}
-          secret-codecov-token: ${{ secrets.CODECOV_TOKEN }}
-          tox-args: '-e napari-dev'
+        shell: bash
+        run: uv pip install git+https://github.com/napari/napari
+
+      - name: Run tests against napari main
```
alessandrofelder, line 107: Just checking: we now lose the code coverage here, right? I think this is fine, because it should be the same as in the other tests.

.github/workflows/test_and_deploy.yml
```diff
@@ -89,21 +89,26 @@ jobs:
...
+        if: github.event_name == 'schedule'
+        shell: bash
+        run: pytest
```
alessandrofelder:
```suggestion
        run: pytest -vv
```
being more verbose could help with the debugging errors caused by changes in the `napari` development version?

alessandrofelder (changes requested):
Thanks @AlgoFoe - a couple of nitpicky comments then this will be good to merge :)

.github/workflows/test_and_deploy.yml
AlgoFoe: Oops, missed that! I'll add the version comment.

.github/workflows/test_and_deploy.yml
AlgoFoe: Yeah, I'll just update the versions to use the SHA in this PR itself.

alessandrofelder (approved):
Nice one, thanks @AlgoFoe

Issue #328, opened 2022-02-11
Remove reliance on `Path.home()` having a specific meaning


alessandrofelder:
Maybe a checkbox called "use pre-trained weights" that disables/enables the textbox to a path for the model depending on whether it's ticked would be better in the Detection widget.

`Path.home` also means "You haven't specified a valid YAML file (or valid YAML files)" where you should have, and "use defaults"  for `model_weights` and `trained_model` in the Training widget.

We should also validate the input of the text box, i.e. check it's the right file format etc. see brainglobe/cellfinder#337 .

Also related brainglobe/cellfinder-napari#43 .

PR #637, opened 2026-07-06
Make the background channel optional in the napari plugin

### LINKED ISSUE #636: [Feature] Choose our "resnet50_1ch" as the default if `background` array is `None`
**Is your feature request related to a problem? Please describe.**
When running on single-channel data wanting to use the default model, and being used to the `cellfinder` Python API, I often make `cellfinder` crash with a dimension mismatch between the model and the data to classify.

**Describe the solution you'd like**
I think it would make sense to use the "default" 1-channel model implicitly if single-channel data is passed (as it's trained on the same training data as the 2-channel model)

aymuos15:
Follow-up to the core single-channel support (#619), extending it through the napari UI so users can run detection, curation, and training with only a signal channel. Also closes #636 by auto-selecting the single-channel model in the Python API.

### What changed
- **Detection** (`detect.py`): the background image is now optional. A missing background is passed to core as `None`. Added a guard: running without a background requires a single-channel model, so the widget prompts the user to enable "Skip classification" or uncheck "Use pre-trained weights" and pick a single-channel model.
- **Curation** (`curation.py`): background layer is optional throughout. When absent, only the signal channel is written, and the training YAML marks `bg_channel = -1` (which core training already reads as signal-only). Signal/background selections reset cleanly to `None` when deselected.
- **Core classify** (`classify.py`): skips background normalization when there is no background array.
- **Core model selection** (`main.py`): when `background_array` is `None` and the default multi-channel model is still selected, the API switches to `resnet50_1ch`, avoiding a dimension mismatch (#636). An explicitly chosen model is left untouched.
- **Containers**: `DataInputs.background_array` is now `Optional`.

### Tests
- Save single-channel training data produces single-channel cubes and a `bg_channel: -1` YAML, verified consumable by core training.
- `check_image_data_for_extraction` accepts signal-only; missing-signal message updated.
- Deselecting the background resets the layer.
- `DataInputs(background_array=None)` passthrough and single-channel detection worker smoke test.
- Missing background auto-selects `resnet50_1ch`; an explicit `resnet50_all` is preserved.

### DIFF
--- a/cellfinder/napari/detect/detect.py
+++ b/cellfinder/napari/detect/detect.py
@@ -261,7 +261,8 @@ def detect_widget() -> FunctionGui:
         soma_spread_factor: float,
         max_cluster_size: float,
         classification_options,
-        model_source: ModelSource,
+        skip_classification: bool,
+        use_pre_trained_weights: bool,
         trained_model: Optional[Path],
         classification_batch_size: int,
         normalize_channels: bool,
@@ -336,12 +337,14 @@ def detect_widget() -> FunctionGui:
             Largest detected cell cluster (in cubic um) where splitting
             should be attempted. Clusters above this size will be labeled
             as artifacts
-        model_source : ModelSource
-            Which classification model to run: the pretrained default, a
-            custom model file, or skip classification entirely
+        skip_classification : bool
+            If selected, the classification step is skipped and all cells from
+            the detection stage are added
+        use_pre_trained_weights : bool
+            Select to use pre-trained model weights
         trained_model : Optional[Path]
-            Trained model file path, used only when model_source is a custom
-            model
+            Trained model file path, used only when pre-trained weights are
+            not selected
         classification_batch_size : int
             How many potential cells to classify at one time. The GPU/CPU
             memory must be able to contain at once this many data cubes for
@@ -396,13 +399,15 @@ def detect_widget() -> FunctionGui:
             show_info("Signal image must be specified.")
             return
 
-        skip_classification = model_source is ModelSource.SKIP
+        model_source = ModelSource.from_options(
+            skip_classification, use_pre_trained_weights
+        )
         if model_source is ModelSource.CUSTOM and not (
             trained_model and Path(trained_model).is_file()
         ):
             show_info(
-                "Select a trained model file, or choose "
-                "'Pretrained default' to use the default model."
+                "Select a trained model file, or check "
+                "'Use pre-trained weights' to use the default model."
             )
             return
 
@@ -450,7 +455,8 @@ def detect_widget() -> FunctionGui:
         )
 
         classification_inputs = ClassificationInputs(
-            model_source,
+            skip_classification,
+            use_pre_trained_weights,
             trained_model,
             classification_batch_size,
             normalize_channels,
--- a/cellfinder/napari/detect/detect.py
+++ b/cellfinder/napari/detect/detect.py
@@ -401,12 +401,13 @@ def detect_widget() -> FunctionGui:
         if (
             options["background_image"] is None
             and not skip_classification
-            and use_pre_trained_weights
+            and not use_pre_trained_weights
         ):
             show_info(
                 "Running without a background image needs a single-channel "
-                "model. Enable 'Skip classification', or uncheck 'Use "
-                "pre-trained weights' and select a single-channel model."
+                "model. Enable 'Use pre-trained weights' to use the default "
+                "single-channel model, enable 'Skip classification', or "
+                "select a single-channel trained model."
             )
             return
 
--- a/cellfinder/napari/detect/detect.py
+++ b/cellfinder/napari/detect/detect.py
@@ -27,6 +27,7 @@ from .detect_containers import (
     DataInputs,
     DetectionInputs,
     MiscInputs,
+    ModelSource,
 )
 from .thread_worker import Worker
 
@@ -260,8 +261,7 @@ def detect_widget() -> FunctionGui:
         soma_spread_factor: float,
         max_cluster_size: float,
         classification_options,
-        skip_classification: bool,
-        use_pre_trained_weights: bool,
+        model_source: ModelSource,
         trained_model: Optional[Path],
         classification_batch_size: int,
         normalize_channels: bool,
@@ -336,14 +336,12 @@ def detect_widget() -> FunctionGui:
             Largest detected cell cluster (in cubic um) where splitting
             should be attempted. Clusters above this size will be labeled
             as artifacts
-        skip_classification : bool
-            If selected, the classification step is skipped and all cells from
-            the detection stage are added
-        use_pre_trained_weights : bool
-            Select to use pre-trained model weights
+        model_source : ModelSource
+            Which classification model to run: the pretrained default, a
+            custom model file, or skip classification entirely
         trained_model : Optional[Path]
-            Trained model file path (home directory (default) -> pretrained
-            weights)
+            Trained model file path, used only when model_source is a custom
+            model
         classification_batch_size : int
             How many potential cells to classify at one time. The GPU/CPU
             memory must be able to contain at once this many data cubes for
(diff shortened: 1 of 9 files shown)


aymuos15:
@alessandrofelder I realize targetting two things at once is slightly messy but they were really intertwined so hence I put it this way. Happy to split your issue seperately if you that makes this revewing easier.

cellfinder/napari/detect/detect.py
```diff
@@ -393,8 +394,20 @@ def widget(
...
+            return
+
+        if (
+            options["background_image"] is None
+            and not skip_classification
+            and use_pre_trained_weights
+        ):
+            show_info(
+                "Running without a background image needs a single-channel "
+                "model. Enable 'Skip classification', or uncheck 'Use "
```
alessandrofelder: I think one should actually _check_ "Use pre-trained weights" to use the default single channel model, no?

(If not, there is no easy way through the GUI to download and use the single-channel default model?)

(and therefore related logic needs to be adapted 🤔 )

alessandrofelder (commented):
Thanks a lot for this @aymuos15 

I've only skimmed the code changes (will come back to it later this week) and done some initial user-testing (largely all looks reasonable to me). Thought it might be useful to flag an initial specific comment before I have time to look more deeply - sorry if I misunderstood!

cellfinder/napari/detect/detect.py
aymuos15: This was a brain fog moment of mine. I think I was just doing everything locally and sided with that. Sorry about that! Thank you very much for catching it! :D

aymuos15: Have commited this.

cellfinder/napari/detect/detect.py
```diff
@@ -393,8 +394,21 @@ def widget(
...
         signal_image = options["signal_image"]
 
-        if signal_image is None or options["background_image"] is None:
-            show_info("Both signal and background images must be specified.")
+        if signal_image is None:
+            show_info("Signal image must be specified.")
+            return
+
+        if (
+            options["background_image"] is None
```
alessandrofelder: sorry - I am still unsure about this 🤔 
Will this not evaluate to `True` and return too early if
* I don't have a background channel
* I want to classify (i.e. I don't want to skip classification)
* I have trained my own single-channel model and therefore do not want to use the pre-trained model?

I wonder whether we want more something like (pseudo-code):
```python
n_channels = 2 if background_image else 1
model_has_correct_number_of channels = use_pre_trained_weights or model.channel_dimension == n_channels

if not model_has_correct_number_of_channels:
     # sensible info message and return
```

aymuos15: Apologies once again, I think I confused myself here. I just wanna quickly make sure my understanding is right this time.

First, for context: **How the default path works (pretrained ticked):**

| Step | File | Action |
|---|---|---|
| 1 | `detect.py:457` | pretrained → `trained_model = None` (`Path.home()` discarded) |
| 2 | `main.py:215` | no background → `resnet50_tv` → `resnet50_1ch` |
| 3 | `main.py:218` → `prep.py:35` | no weights given → `download_models(...)` |
| 4 | `tools.py:get_model` | no custom model → `build_model` + `load_weights(default)` |
| 5 | `classify.py:180` | channel check |

This guard was set up because the tick-box `use_pre_trained_weights` is the real switch, but `trained_model` starts as `Path.home()` (a folder), not empty, and only becomes empty when the tick-box is on. So when the tick-box is off, there's no clean way to tell "I picked a model" from "I picked nothing", both look like a path. The guard just patches over that. I think this was what led to my choice above. 

Maybe, we can actually just collapse to pretrained / custom / skip into one exclusive choice, so "nothing picked" is a checkable state and the guard isn't needed. Current vs proposed behaviour:

| Model choice | Background | Current | Proposed |
|---|---|---|---|
| Pretrained | yes | 2ch default, proceeds | same |
| Pretrained | no | 1ch default, proceeds | same |
| Custom file | yes | proceeds; core errors if ≠ 2ch | same |
| Custom file | no | **guard blocks it** | proceeds; core errors if ≠ 1ch |
| No model picked | yes | cryptic `load_model(home)` error | blocked, "select a file" |
| No model picked | no | guard blocks (masks that error) | blocked, "select a file" |
| Skip classification | yes | detection only | same |
| Skip classification | no | detection only | same |

Since the logic you mentioned already exists there (`classify.py:180-188`), can we just follow the above and patch it there?

```python
# detect_containers.py
class ModelSource(Enum):
    PRETRAINED = "Pretrained default"
    CUSTOM = "Custom model"
    SKIP = "Skip classification"

# ClassificationInputs: one field replaces the two flags
    model_source: ModelSource = ModelSource.PRETRAINED

# as_core_arguments(): translate the choice for the core
    args["skip_classification"] = self.model_source is ModelSource.SKIP
    if self.model_source is not ModelSource.CUSTOM:
        args["trained_model"] = None   # core downloads the default
```

```python
# detect.py: derive skip, guard only the "custom but no file" case
    skip_classification = model_source is ModelSource.SKIP
    if model_source is ModelSource.CUSTOM and not (
        trained_model and Path(trained_model).is_file()
    ):
        show_info("Select a trained model file.")
        return
```

alessandrofelder: No worries at all, and thanks for dissecting this @aymuos15 - the fact that we use both a Path and a checkbox to encode one thing is confusing currently for sure! (See related 4-year issue I made that never went anywhere 😂 https://github.com/brainglobe/cellfinder/issues/328)

I agree with the "proposed" column in your table above, I think that's what the behaviour should be. My only minor note would be that the error message should also not the option to tick "pretrained"?
```python
# detect.py: derive skip, guard only the "custom but no file" case
    skip_classification = model_source is ModelSource.SKIP
    if model_source is ModelSource.CUSTOM and not (
        trained_model and Path(trained_model).is_file()
    ):
        show_info("Select a trained model file or check the `Use pre-trained weights` to use the default model.")
        return
  ```

Comment (aymuos15) [reply cellfinder/napari/detect/detect.py]: My bad, Shouldve spotted that issue in advance haha. Thank you for pointing it out, I have added it to the todo list in our projct and will PR that after this. 

In my latest commit, I have done what we discused in the table along with your proposed error message. Thank you!

**Comment by alessandrofelder** [comment]
Thanks @aymuos15 - I think we are almost there, but I misunderstood some of your intentions - sorry! Thanks for your patience!

I like the `ModelSource` enum under the hood for code legibility a lot (and the logic is now sound and I can use the single-channel model via napari :tada: ), but on the user-facing side, I would like to keep two checkboxes because I think they are the natural UI element for `skip` and `use-pretrained`. 

Ideally, I would like to add an interactive element where ticking `use-pretrained` hides the textfield for the custom model? (I think this should be a separate issue though!) What do you think?

**Comment by aymuos15** [comment]
Really glad it's working now! Sorry it took so many iterations.

Regarding the UI, can you check this diff then, is this what you had in mind? Two checkboxes back on the front, `ModelSource` kept under the hood and derived from them:

```python
class ModelSource(Enum):
    PRETRAINED = auto()
    CUSTOM = auto()
    SKIP = auto()

    @classmethod
    def from_options(
        cls, skip_classification: bool, use_pre_trained_weights: bool
    ) -> "ModelSource":
        if skip_classification:
            return cls.SKIP
        return cls.PRETRAINED if use_pre_trained_weights else cls.CUSTOM
```

Both boxes can be ticked at once now, so skip wins; there's a test over all four combinations.

I think the custom model text field is a small change so maybe I can fold it in here?

```python
def update_trained_model_visibility() -> None:
    widget.trained_model.visible = (
        ModelSource.from_options(
            widget.skip_classification.value,
            widget.use_pre_trained_weights.value,
        )
        is ModelSource.CUSTOM
    )

widget.skip_classification.changed.connect(update_trained_model_visibility)
widget.use_pre_trained_weights.changed.connect(
    update_trained_model_visibility
)
update_trained_model_visibility()
```

It's a separate commit, so easy to drop if you'd rather it stay its own issue.

alessandrofelder:
Happy for it to be folded in here if you are.

>is this what you had in mind? 

Yes, exactly, based on your comment (don't think you pushed any new commits yet).

aymuos15:
Pushed now, just wanted to confirm before committing.

alessandrofelder (approved):
Thanks for your patience with the back and forth, and thanks for this @aymuos15 .
I am happy with it now, we can continue to explore improvements to the single channel model - but that's orthogonal to this PR.

aymuos15:
Thank you very much for the guidance!

Issue #298, opened 2021-06-29
2D data support


adamltyson:
General issue to track progress towards 2.5D data (i.e. serial 2D sections, where axial sampling is much larger than cell soma diameter) support.

To be done:

- [ ] Implement 2D ResNet, adapting [existing code](https://github.com/brainglobe/cellfinder-core/blob/45655751419c288201e51cde2c2a2833f81450c7/cellfinder_core/classify/resnet.py#L36). Should be changing input sizes, kernel sizes, and using 2D keras layers.
- [ ]  Implement a 2D [CubeGenerator](https://github.com/brainglobe/cellfinder-core/blob/45655751419c288201e51cde2c2a2833f81450c7/cellfinder_core/classify/cube_generator.py#L21)
- [ ] Retrain 3D network with 2D data, e.g. using central N planes of the 3D data detailed [here](https://docs.brainglobe.info/cellfinder/user-guide/training/using-supplied-data) ([direct link to data](https://gin.g-node.org/cellfinder/training_data/raw/master/serial2p.tar.gz)).
- [ ]  Implement [existing 2D cell candidate detection](https://github.com/brainglobe/cellfinder-core/blob/45655751419c288201e51cde2c2a2833f81450c7/cellfinder_core/detect/filters/plane/multiprocessing.py#L9) on 2D data.
- [ ] Implement 2D filtering of cell candidate pixels from initial detection. Could implement a 2D version of [existing 3D code](https://github.com/brainglobe/cellfinder-core/blob/main/cellfinder_core/detect/filters/volume/multiprocessing.py), but a simpler scikit-image/scipy implementation may be quick enough for 2.5D data (e.g. 1/10th data volume, and only 2D processing). 
- [ ] Support both 2/3D inputs. Do we decide when to use each (based on axial sampling and soma size), or does the user choose? Maybe the software decides, but with an override flag.
- [ ] Refactor code to support 2/3D throughout (i.e. single functions, and no code duplication).
- [ ] Support upstream software ([cellfinder](https://github.com/brainglobe/cellfinder/issues/79) & [cellfinder-napari](https://github.com/brainglobe/cellfinder/issues/79)).

PR #490, opened 2025-03-07
Inform the user when they supply just the weights in place of the full model

parharti:
This PR improves the error handling when users provide only model weights instead of the full model in Cellfinder. Previously, users would receive a generic shape mismatch error, which did not clearly explain the issue or suggest a solution. This update enhances the error message, providing users with clearer guidance on resolving the problem.
#454

PR #643, opened 2026-07-16
Support 2D classification

aymuos15:
Adds 2D classification via a depth-1 cube. A helper validates the 2D/3D processing mode, and the ResNet classifier accepts both 2D and 3D inputs. Standalone: detection, training, and napari support follow in later PRs of this stack.

**Stack:** first of six.

Progresses #298 (2D ResNet and 2D CubeGenerator).

## Changes
- Accept 2D and 3D inputs in the ResNet classifier.
- Add a helper to validate the 2D/3D processing mode.
- Support classification on a depth-1 cube.
- Unit tests for the classifier build, the mode helper, and cube generation.

## Design
- One ResNet builder resolves the convolution and pooling rank from the input shape, so the same code emits a 2D or 3D network with no duplicated builder.
- Classification reuses the existing cube generator with a depth-1 cube and squeezes the z axis; it accepts 2D or 3D signal/background arrays.
- `dimensions` defaults to 3, so no existing 3D behaviour changes at the defaults.

## Testing
Unit tests cover the added lines: the 2D and 3D classifier builds, the mode validator, and depth-1 cube generation. Run under `KERAS_BACKEND=torch`.

## Results

ResNet-50, 100 epochs, 10,756 held-out test cubes.

| Model | Test acc | F1 | Precision | Recall | ROC-AUC | AP |
|-------|----------|-----|-----------|--------|---------|-----|
| 3D 2-channel (existing default) | 0.979 | 0.978 | 0.971 | 0.984 | 0.997 | 0.995 |
| 2D 2-channel | 0.974 | 0.973 | 0.953 | 0.994 | 0.995 | 0.993 |
| 2D 1-channel | 0.964 | 0.963 | 0.951 | 0.974 | 0.992 | 0.989 |

The 3D row is the existing default model, for reference; 2D trades a little accuracy for the reduced input. Nothing in this PR changes the 3D path.

### Training curves
2ch training curves
1ch training curves

### Confusion matrices
2ch confusion
1ch confusion

### DIFF
--- a/cellfinder/core/classify/classify.py
+++ b/cellfinder/core/classify/classify.py
@@ -92,15 +92,16 @@ def _squeeze_depth1_batch(data, z_axis: int):
 def _run_inference(
     model,
     data_loader: DataLoader,
-    sampler: CuboidBatchSampler,
     dataset: CuboidArrayDataset,
     workers: int,
-    dimensions: int,
-    z_axis: int,
+    squeeze_z_axis: Optional[int],
     callback: Optional[Callable[[int], None]],
 ) -> np.ndarray:
     """
     Run the model over every batch, returning the concatenated outputs.
+
+    If `squeeze_z_axis` is not None, each batch has that axis dropped to give
+    2D network input.
     """
     if workers:
         dataset.start_dataset_thread(workers)
@@ -108,14 +109,14 @@ def _run_inference(
         outputs = []
         for step, data in tqdm.tqdm(
             enumerate(data_loader),
-            total=len(sampler),
+            total=len(data_loader.sampler),
             desc="Classifying",
             unit="batches",
             smoothing=1 / (25 * max(1, workers)),
             mininterval=0.5,
         ):
-            if dimensions == 2:
-                data = _squeeze_depth1_batch(data, z_axis)
+            if squeeze_z_axis is not None:
+                data = _squeeze_depth1_batch(data, squeeze_z_axis)
             output = model(data)
             # in original keras, it seemed to held on to the output until the
             # end (possibly on GPU). This causes resources issues for very
@@ -319,11 +320,13 @@ def main(
     outputs = _run_inference(
         model=model,
         data_loader=data_loader,
-        sampler=sampler,
         dataset=dataset,
         workers=workers,
-        dimensions=dimensions,
-        z_axis=dataset.output_axis_order.index("z") + 1,
+        squeeze_z_axis=(
+            dataset.output_axis_order.index("z") + 1
+            if dimensions == 2
+            else None
+        ),
         callback=callback,
     )
 
--- a/cellfinder/core/classify/classify.py
+++ b/cellfinder/core/classify/classify.py
@@ -17,7 +17,10 @@ from cellfinder.core.classify.cube_generator import (
 )
 from cellfinder.core.classify.tools import get_model, model_input_channels
 from cellfinder.core.tools.image_processing import dataset_mean_std
-from cellfinder.core.tools.tools import deprecate_positional_args
+from cellfinder.core.tools.tools import (
+    deprecate_positional_args,
+    validate_dimensions,
+)
 from cellfinder.core.train.train_yaml import depth_type, models
 
 
@@ -39,6 +42,7 @@ def main(
     network_depth: depth_type,
     max_workers: int = 3,
     pin_memory: bool = False,
+    dimensions: int = 3,
     callback: Optional[Callable[[int], None]] = None,
     normalize_channels: bool = False,
     normalization_n_sampling_planes: int = 50,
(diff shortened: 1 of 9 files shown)


cellfinder/core/classify/classify.py
```diff
@@ -106,10 +220,20 @@ def main(
...
+    else:
+        (
+            signal_array,
+            background_array,
+            voxel_sizes,
+            network_voxel_sizes,
+        ) = _promote_2d_arrays(
+            signal_array, background_array, voxel_sizes, network_voxel_sizes
+        )
+        cube_depth = 1
```
alessandrofelder: Would it be better to promote first and then validate?
Currently bugs introduced by changes to `_promote_2d_arrays` would not be caught by validation?

cellfinder/core/classify/tools.py
```diff
@@ -13,6 +13,52 @@ def model_input_channels(model: Model) -> int:
...
+    logger.debug(f"Setting model weights according to: {model_weights}")
+    if model_weights is None:
+        raise OSError(
+            "`model_weights` must be provided for inference "
+            "or continued training."
+        )
+    try:
+        model.optimizer.build(model.trainable_variables)
+        model.load_weights(model_weights)
+    except (OSError, ValueError) as e:
```
alessandrofelder, line 55: Out of curiousity, why do we also catch an `OSError` here?

cellfinder/core/classify/tools.py
```diff
@@ -13,6 +13,52 @@ def model_input_channels(model: Model) -> int:
     return tuple(model.inputs[0].shape)[-1]
 
 
+def _load_existing_model(
+    existing_model: Optional[os.PathLike], dimensions: int
+) -> Model:
+    """
+    Load a saved model, checking its input rank matches `dimensions`.
```
alessandrofelder, line 20: It seems later on in the code, we might call this with `existing_model=None`. If this is intended, could we document what `keras.models.load_model(None)` returns?

cellfinder/core/classify/tools.py
```diff
@@ -36,37 +84,32 @@ def get_model(
...
+    :param shape: The input shape (excluding the batch dimension) to use when
+    creating a new model. If None, a default for `dimensions` with
+    `num_channels` channels is used.
     :return: A keras model
 
     """
     if existing_model is not None or network_depth is None:
-        logger.debug(f"Loading model: {existing_model}")
-        model = keras.models.load_model(existing_model)
+        model = _load_existing_model(existing_model, dimensions)
```
alessandrofelder, line 103: See comment above: what happens when `existing_model` and `network_depth` are both `None`?

alessandrofelder (changes requested):
Thanks a lot @aymuos15 - the code changes look promising... I left some tiny comments/questions to satisfy my curiousity/for your consideration.

I'll do some user-testing next week (do I need a specific new model from you for this?)

cellfinder/core/classify/tools.py
aymuos15: This is pre-existing, from #490 (489d97b), just moved into the helper here:

```diff
+            try:
+                model.load_weights(model_weights)
+            except (OSError, ValueError) as e:
+                raise ValueError(
+                    f"Error loading weights: {model_weights}.\n"
+                    "Provided weights don't match the model architecture.\n"
+                ) from e
```

Both arms are reachable from `load_weights`: on keras 3.15.1 a missing path raises `FileNotFoundError` and a corrupt file `OSError`, while an architecture mismatch raises `ValueError` (the case `test_incorrect_weights` covers). Without the `OSError` - a mistyped path escapes as a raw h5py `[Errno 2] Unable to synchronously open file`.

cellfinder/core/classify/tools.py
aymuos15: Thank you for raising this. `load_model(None)` raises `TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType` from inside keras, so the both-`None` case died with an opaque error. Guarded explicitly in e606b94:

```python
if existing_model is not None:
    model = _load_existing_model(existing_model, dimensions)
elif network_depth is None:
    raise ValueError(
        "Either `existing_model` or `network_depth` must be provided."
    )
```

Both in-tree callers pass `network_depth=models[...]`, so it was only reachable by calling `get_model` directly. Test added. This supersedes #593. So I guess we can close this now?

cellfinder/core/classify/tools.py
aymuos15: Guarded in e606b94, so `_load_existing_model` can no longer receive `None`. `Optional` dropped from its signature.

cellfinder/core/classify/classify.py
aymuos15: Thank you again for raising this. Moved the validation to run after promotion, on both paths, in af708aa:

```python
if dimensions == 2:
    (...) = _promote_2d_arrays(...)
    cube_depth = 1

_validate_3d_arrays(signal_array, background_array)
```

`_promote_2d_arrays` keeps its own 2D/3D input check so that case still gets a specific message, but whatever it returns now has to be 3D before it reaches the cube generator.

alessandrofelder (approved):
Thanks a lot @aymuos15 - looking forward to the follow up PRs to this 🎉

aymuos15:
Thank you so much!!

PR #646, opened 2026-07-31
Use single underscore for internal CurationWidget methods

### LINKED ISSUE #624: CurationWidget uses double-underscore names for internal methods that should be single underscore
**Describe the problem**

Five internal methods on `CurationWidget` in `cellfinder/napari/curation.py` are named with a leading double underscore:

- `__prep_directories_for_save`
- `__delete_existing_saved_training_data`
- `__extract_cubes`
- `__finish_save`
- `__save_yaml_file`

A leading `__` triggers Python name mangling: the identifier is rewritten to `_CurationWidget__name` at compile time. Per the [language reference](https://docs.python.org/3/reference/expressions.html#index-5) and [tutorial](https://docs.python.org/3/tutorial/classes.html#private-variables), mangling exists specifically to avoid name clashes with subclasses, not to mark a method as internal. The convention for "non-public, treat as internal" is a single leading underscore. These methods are plain internal helpers and are not designed to resist subclass overrides.

**Expected**

Internal helper methods use a single leading underscore, consistent with the rest of the codebase.

**Additional context**

This is the established style everywhere else in the repo. Repo-wide there are 40 single-underscore internal methods across 12 files versus these 5 double-underscore-as-internal methods, all confined to `curation.py`. The only other dunder method defs in the codebase are genuine Python protocol methods (`__init__`, `__call__`, `__len__`, `__getitem__`, `__iter__`, `__getstate__`), which are correct. `curation.py` is itself mixed: it already has single-underscore helpers like `_add_training_data_layers` alongside the five dunders.

This was raised by @alessandrofelder in review of PR #621, where the new `__finish_save` method followed the file's existing dunder style for consistency. The other four predate it.

**Proposed fix**

Rename the five methods to single underscore and update their call sites within `CurationWidget`:

| Current | Renamed to |
| --- | --- |
| `__prep_directories_for_save` | `_prep_directories_for_save` |
| `__delete_existing_saved_training_data` | `_delete_existing_saved_training_data` |
| `__extract_cubes` | `_extract_cubes` |
| `__finish_save` | `_finish_save` |
| `__save_yaml_file` | `_save_yaml_file` |

All references are internal to the class, so no public API changes.

aymuos15:
Closes #624

## Description

Renames five internal helpers in `cellfinder/napari/curation.py` from `__name` to `_name`. A leading `__` triggers name mangling, which exists to avoid clashes with subclasses rather than to mark a method as internal; the convention for internal is a single underscore, as used everywhere else in the repo and by existing helpers in this same file.

| Current | Renamed to |
| --- | --- |
| `__prep_directories_for_save` | `_prep_directories_for_save` |
| `__delete_existing_saved_training_data` | `_delete_existing_saved_training_data` |
| `__extract_cubes` | `_extract_cubes` |
| `__finish_save` | `_finish_save` |
| `__save_yaml_file` | `_save_yaml_file` |

Raised by @alessandrofelder in review of #621.

## What this does not change

Genuine Python protocol methods (`__init__`, `__call__`, `__len__`, etc.) elsewhere in the codebase are untouched.

## References

All call sites are internal to `CurationWidget`, so there is no public API change and no behaviour change. Pure rename: 11 lines, all in `cellfinder/napari/curation.py`.

### DIFF
--- a/cellfinder/napari/curation.py
+++ b/cellfinder/napari/curation.py
@@ -489,19 +489,19 @@ class CurationWidget(QWidget):
                     if not choice:
                         return
             if self.output_directory is not None:
-                self.__prep_directories_for_save()
-                self.__extract_cubes(block=block)
+                self._prep_directories_for_save()
+                self._extract_cubes(block=block)
                 if block:
-                    self.__finish_save()
+                    self._finish_save()
 
-    def __prep_directories_for_save(self):
+    def _prep_directories_for_save(self):
         self.yaml_filename = self.output_directory / "training.yaml"
         self.cell_cube_dir = self.output_directory / "cells"
         self.no_cell_cube_dir = self.output_directory / "non_cells"
 
-        self.__delete_existing_saved_training_data()
+        self._delete_existing_saved_training_data()
 
-    def __delete_existing_saved_training_data(self):
+    def _delete_existing_saved_training_data(self):
         self.yaml_filename.unlink(missing_ok=True)
         for directory in (
             self.cell_cube_dir,
@@ -512,7 +512,7 @@ class CurationWidget(QWidget):
             else:
                 directory.mkdir(exist_ok=True, parents=True)
 
-    def __extract_cubes(self, *, block=False):
+    def _extract_cubes(self, *, block=False):
         """
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


alessandrofelder (approved):
Thanks @aymuos15

